Issue #11, opened 2017-10-23
[PagedPublicationKit]  Enhance `beforeNavigation`/`beforeNavigation` event data


klarstrup:
It currently takes too much work to figure out which *pages*(not spreads) we are coming from/to.

Issue #10, opened 2017-10-23
[PagedPublicationKit] Provide events to handle the results of catalog, offer, hotspot fetching


klarstrup:
This lets anyone using PagedPublicationKit utilize the data fetched for their own use.

We need this for offline support.

Issue #8, opened 2017-10-23
[PagedPublicationKit] Let be primed with catalogs/pages/hotspots data


klarstrup:
We need this for offline support.

PR #12, opened 2017-10-23
#8 #10 #11 and read progress tweak and isbrowser fix

klarstrup:


lib/coffeescript/kits/paged-publication/core.coffee
```diff
@@ -172,10 +172,12 @@ class PagedPublicationCore
 
     beforeNavigation: (e) ->
         position = e.newPosition
-        versoPageSpread = @getVerso().getPageSpreadFromPosition position
+        theVerso = @getVerso()
```
mortenbo: Odd naming IMO.

lib/coffeescript/kits/paged-publication/initialize.coffee
```diff
@@ -87,6 +96,12 @@ module.exports = (options = {}, callback) ->
             return
 
         callback null, viewer, data
+        if !options.data.details
+            viewer.trigger 'catalogFetched', data.details
```
mortenbo: We should never use the word `catalog`. From now on, it's paged publication. The only place where we have catalog is in the actual API URL's.

lib/coffeescript/kits/paged-publication/core.coffee
klarstrup: `verso` key was already taken up by the event info, what's another alternative? `versoInstance`?

mortenbo: The variable 'verso' isn't used in this function as far as I can see?

klarstrup: No but that name is used for a key on the `beforeNavigation` event object and I wanted to avoid any confusion

mortenbo: Can't you use e.verso instead of @getVerso() then?

klarstrup: Are you suggesting we include the verso instance in the events from `verso-browser`? It's not right now in any case, might be widely useful? https://github.com/shopgun/verso-browser/blob/develop/lib/coffeescript/verso.coffee#L119

mortenbo: Ah, right. What do you think? Might be a tad nasty with instances in event data but we already do it with other instances so might not be an issue. Otherwise, `theVerso` is fine for now.

klarstrup: Can we just uniformly have the verso instance be the second argument to all events from verso? Does microevent let us do that?

Feel free to merge and maybe create an issue over on shopgun/verso-browser

PR #23, opened 2017-12-04
Expose kits/asset as AssetKit

izotos:


klarstrup:
Doesn't break reactlab so I'm sure it's fine. 👍

izotos:
#24 Sorry about that, renamed for consistency

PR #31, opened 2018-03-19
Unify entries

klarstrup:
# 🤓


klarstrup:
Oh yeah, we can get rid of that.

Issue #5, opened 2017-09-28
Add Promise support to .request methods


klarstrup:
Let the `.request` methods of the graphkit and the corekit return a `Promise` when not passed a callback function in the last argument, so as to provide an API like so:
```js
SGN.CoreKit.request({ url: '/v2/catalogs' })
    .then(function(data){/* handle data */})
    .catch(function(error){/* handle error */})
```

Check the environment for support and throw an error if support is not present. Said error should suggest using the callback style or adding a `Promise` polyfill.

mortenbo:
@klarstrup I was wondering if this should be a responsibility of the SDK. If yes, we need to bundle Promise support into the SDK for browser environments, which will increase the bundle size.

klarstrup:
It's easier to let end users(ie. app developers) decide whether they need to support browsers without native `Promise` support and set up the global polyfill if necessary. From our perspective it doesn't make a difference where the support comes from. 

But we should definitely give an error message if they don't provide a callback function and we don't find `Promise` support, something presenting their options: Using the callback style or providing Promise support.

I will try to find a way to easily apply this pattern to all asynchronous methods throughout the kit.

mortenbo:
@klarstrup We should act on this soon or close it, IMO. Perhaps it can do done like AWS S3 where you can call `.promise()` at the end of e.g. `getObject()`.

PR #43, opened 2018-06-04
add promise interops to graphkit and corekit request methods #5

klarstrup:


klarstrup:
S+M whenever you've had a look and feel like it @mortenbo

__tests__/core_kit.spec.coffee
```diff
@@ -19,6 +19,11 @@ test 'Making a request with JSON response', (done) ->
 
     return
 
+test 'Making a request with JSON response, promises async/await style', ->
+    data = await SGN.CoreKit.request url: '/v2/catalogs'
+    expect(data).toBeDefined()
```
mortenbo: I think we should have spacing between function calls and variable declarations. I don't know what the Coffee rule is but eslint rule is: `"keyword-spacing": ["error", { "before": true, "after": true }],`.

__tests__/graph_kit.spec.coffee
```diff
@@ -19,3 +19,16 @@ test 'Making a request', (done) ->
...
+    data = await SGN.GraphKit.request
+        query: """query PlanningSuggestions {
+          planningSuggestions(locale: "da_DK", term: "c") {
+            phrases {
+              name
+            }
+          }
+        }
+        """,
+        operationName: 'PlanningSuggestions'
```
mortenbo: Same here.

__tests__/util_promiseCallbackInterop.coffee
```diff
@@ -0,0 +1,125 @@
+{ promiseCallbackInterop } = require '../lib/coffeescript/util'
+testResult = { results: ['a result'] }
+testError = { error: 'messed up' }
```
mortenbo: I think the default ES rule is no spacing between {}.

__tests__/util_promiseCallbackInterop.coffee
```diff
@@ -0,0 +1,125 @@
...
+testFun1 = (callback) ->
+    setTimeout(->
+        callback(undefined, testResult)
+    )
+testFun1Error = (callback) ->
+    setTimeout(->
```
mortenbo: Why paranthesis? I would do:

setTimeout ->
    callback undefined, testResult

    return

I think it's important we follow the style of the code in the project instead of having multiple code styles.

lib/coffeescript/kits/core/request.coffee
```diff
@@ -65,4 +66,4 @@ request = (options = {}, callback = ->, runs = 0) ->
 
     return
 
-module.exports = request
+module.exports = promiseCallbackInterop(request, 1)
```
mortenbo: We don't use paranthesis around funs anywhere else in the project.

lib/coffeescript/util.coffee
```diff
@@ -233,5 +233,41 @@ util =
                 k++
 
             return
+    # method for wrapping a function that takes a callback in any position
```
mortenbo: Small stuff: I think we should use correct capitalization rules, periods, etc.

klarstrup:
I'll fix those up now, but going forward it would be nicer if we can have styling enforced either by an autoformatter or a linter. It's a waste of time for you to point out and for me to fix.

Sadly right now coffeelint does not support the rules we'd need: Mandatory parentheses omission on function call parameter lists, mandatory braces omission on objects. (But it does support `no_implicit_braces` and `no_implicit_parens` rules enforcing the opposite. 🤔)

mortenbo:
> I'll fix those up now, but going forward it would be nicer if we can have styling enforced by an autoformatter and/or a linter. It's a waste of time for you to point out and for me to fix.
> 
> Sadly right now coffeelint does not support the rules we'd need: Mandatory parentheses omission on function call parameter lists, mandatory braces omission on objects. (But it does support no_implicit_braces and no_implicit_parens rules enforcing the opposite. 🤔)

Yeah, the linter really lacks a bunch of rules. Looks good now! Feel free to merge it.

klarstrup:
👌

PR #52, opened 2018-08-27
Add trackEvent type 6 as per SGN-110

klarstrup:


lib/coffeescript/kits/events/tracker.coffee
```diff
@@ -152,6 +153,10 @@ module.exports = class Tracker
     
     trackSearched: (properties, version) ->
         @trackEvent 5, properties, version
+
+    # 6 to be used as detailed in SGN-110
```
mortenbo: @klarstrup You are referencing an internal name in a public repo.

PR #57, opened 2018-10-01
Re-add JSON body behavior to CoreKit request, fixes #56

### LINKED ISSUE #56: CoreKit.request `body` behavior
Prior to 2.2.11 SGN.CoreKit.request would assume any provided request `body` should be JSON encoded and sent off with JSON mime headers: https://github.com/shopgun/shopgun-js-sdk/blob/dad037561f49c3c18b140b764a0d5e66bef825aa/lib/coffeescript/kits/core/request.coffee#L10

As of 2.2.11 this is no longer the case and currently it is necessary to write out those things:
```js
    const session = await SGN.CoreKit.request({
      url: "/v2/sessions",
      method: "PUT",
      headers: {
        Accept: "application/json",
        "Content-Type": "application/json",
      },
      body: JSON.stringify({
        email,
        password,
      }),
    });
```

I'm going to restore this behavior as well as reintroduce the "json" option.

klarstrup:
fixes #56

### DIFF
--- a/lib/coffeescript/kits/core/request.coffee
+++ b/lib/coffeescript/kits/core/request.coffee
@@ -8,6 +8,7 @@ request = (options = {}, callback, secondTime) ->
 
         url = SGN.config.get('coreUrl') + (options.url ? '')
         headers = options.headers ? {}
+        json = if typeof options.json is 'boolean' then options.json else true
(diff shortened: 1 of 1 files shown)


klarstrup:
Alright, I cleaned it up a bit :)

PR #60, opened 2019-03-07
Event 8: Incito Publication Opened support in EventTracker & IncitoPublicationKit

klarstrup:


lib/coffeescript/kits/incito-publication/event-tracking.coffee
```diff
@@ -0,0 +1,21 @@
+MicroEvent = require 'microevent'
+
+class IncitoPublicationEventTracking
+    constructor: (@eventTracker, @id, {
+        @pagedPublicationId
```
mortenbo, line 5: Didn't know you could expand like this in CoffeeScript. Neat.

dist/js/incito-publication.js
```diff
@@ -4,6 +4,9 @@
             graphUrl: 'https://graph.service-staging.shopgun.com'
         });
     }
+    SGN.config.set({
```
mortenbo: Perhaps you could remove all the staging configs as we don't really want them here.

mortenbo (approved):
Looks very good but I think we should remove the service-staging config in dist/js/incito-publication.js. Awesome job!

PR #64, opened 2019-12-03
Let CoreUIKit OfferDetails grow beyond anchor

klarstrup:
Let CoreUIKit OfferDetails grow beyond the width of its anchor, with respect to on-screen position.

Also adds options to the `OfferDetails` constructor to allow fine-tuning positioning:
`minWidth`: Default `300px` or `anchorEl` width, whichever is larger(so `OfferDetails` will never be narrower than `anchorEl`)
 `maxWidth`: Default `100vw`

Motivation: https://3.basecamp.com/4286639/buckets/13812214/todos/2144440834

### DIFF
--- a/lib/coffeescript/kits/core-ui/offer-details.coffee
+++ b/lib/coffeescript/kits/core-ui/offer-details.coffee
@@ -1,10 +1,14 @@
 module.exports = class OfferDetails
-    constructor: (@options = {}) ->
+    constructor: ({ @minWidth = 300, @maxWidth = "100vw", @anchorEl, @contentEl }) ->
+        @elInner = document.createElement 'div'
+        @elInner.className = 'sgn-offer-details-inner'
+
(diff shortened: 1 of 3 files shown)


lib/coffeescript/kits/core-ui/offer-details.coffee
```diff
@@ -1,10 +1,14 @@
 module.exports = class OfferDetails
-    constructor: (@options = {}) ->
+    constructor: ({ @minWidth = 300, @maxWidth = "100vw", @anchorEl, @contentEl }) ->
```
mortenbo: Please use single apostrophes.

mortenbo: And no space between {}

mortenbo (changes requested):

PR #73, opened 2020-09-26
Added punctuation :)

Zerplin:


mortenbo (changes requested):
Thanks for the pull request. We'd like to see a bit more improvements before adding this in ;-)

PR #74, opened 2020-10-12
PagedPublicationKit: Desktop high res page images

klarstrup:
PagedPublicationKit will now use zoom-res images for pages prior to zooming in, when the physical pixel width of the page would be over 700 pixels wide.

### DIFF
--- a/lib/coffeescript/kits/paged-publication/page-spread.coffee
+++ b/lib/coffeescript/kits/paged-publication/page-spread.coffee
@@ -41,7 +41,14 @@ class PagedPublicationPageSpread
         imageLoads = 0
 
         pages.forEach (page, i) =>
+            maxPageWidth = el.clientWidth * (window.devicePixelRatio or 1)
+            if @options.pageMode is "double"
+                maxPageWidth = maxPageWidth / 2
+
             image = page.images.medium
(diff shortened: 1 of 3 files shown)


lib/coffeescript/kits/paged-publication/page-spread.coffee
```diff
@@ -41,7 +41,14 @@ class PagedPublicationPageSpread
         imageLoads = 0
 
         pages.forEach (page, i) =>
+            maxPageWidth = el.clientWidth * (window.devicePixelRatio or 1)
```
mortenbo: I think this operation should be moved outside of the loop to avoid calculating it over and over again. Don't know what the hit is here but grabbing dimensions from elements is usually expensive. So outside of the loop we could have an `useLargeImage = true | false` and then do `image = if useLargeImage then page.images.larse else page.images.medium`

klarstrup: It's a good point, I had confused myself into thinking pageMode could change per-page. Fixed.

But the hit is either one or none extra as a spread can have up to 2 pages, I think. :P

mortenbo (approved):

mortenbo (approved):

PR #80, opened 2021-03-09
feat: add html-view to incito-browser

martensievers:
This PR adds support for a basic `HTMLView` into the `incito-browser` library, which supports an additional style property that allows the use of valid CSS style syntax.

By default an `HTMLView` is a `div` element with custom style properties.

### DIFF
--- a/lib/incito-browser/views/html-view.styl
+++ b/lib/incito-browser/views/html-view.styl
@@ -1,0 +1,2 @@
+.incito__html-view
+  margin: 0
--- a/lib/incito-browser/views/html-view.styl
+++ b/lib/incito-browser/views/html-view.styl
@@ -1,1 +1,0 @@
-.incito__html-view
--- a/lib/incito-browser/views/html-view.styl
+++ b/lib/incito-browser/views/html-view.styl
@@ -1,2 +1,1 @@
 .incito__html-view
-  margin: 0
--- a/lib/incito-browser/views/html-view.js
+++ b/lib/incito-browser/views/html-view.js
@@ -1,0 +1,53 @@
+import {isDefinedStr} from '../utils';
+import './html-view.styl';
+import View from './view';
+
+
+class HTMLView extends View {
+    render() {
+        if (isDefinedStr(this.attrs.style)) {
+            this.el.style = this.attrs.style;
+        }
+
+        if (isDefinedStr(this.attrs.text)) {
+            this.el.innerText = this.attrs.text;
+        }
+
+        if (isDefinedStr(this.attrs.src)) {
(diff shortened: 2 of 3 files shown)


mortenbo:
Elegant solution. Would it make sense to only have `div` and `img` as options?

lib/incito-browser/views/html-view.styl
```diff
@@ -0,0 +1,2 @@
+.incito__html-view
+  margin: 0
```
mortenbo: .incito__view already has margin: 0;

klarstrup:
Just skip the `styl` file entirely for not if it's not applicable. It's also not valid stylus code to have a selector without any declarations.

klarstrup (changes requested):
Just skip the `styl` file entirely for now if it's not applicable. It's also not valid stylus code to have a selector without any declarations, so the build is broken atm.

One thought I had, if HTMLView only exists to allow arbitrary `style` - another option it to allow `style` on the base `View`s?

martensievers:
> Just skip the styl file entirely for not if it's not applicable. It's also not valid stylus code to have a selector without any declarations, so the build is broken atm.
> 
> One thought I had, if HTMLView only exists to allow arbitrary `style` - another option it to allow `style` on the base `View`s?

I guess that could theoretically work as well, however, might not be as clear and distinct maybe? @mortenbo any thoughts on this?

klarstrup:
I think in a lot of ways this change - however we do it - is a philosophical fork in the road for Incito, yeah.

klarstrup (approved):

martensievers:
You're right - it's definitely going against the initial idea of having a strictly defined format that can only do things that work across all platforms - but we have come to a point where we no longer have SDK's for all platforms and our functionality is very much limited in contrast to what our competitors can do if they just use HTML + CSS so I think it's maybe an evolutionary step for the format to go in this direction. But if we do it as part of the Incito format, then Incito can still decide which technologies we adapt and utilize and we can limit certain CSS styles if they aren't cross-browser compatible for example.

martensievers:
@mortenbo do you think we should add the `HTMLView` or add a `style` property to the core `View`s?

mortenbo:
I think a `HTMLDivView` makes sense. It's just a type of view like we have a VideoView.

mortenbo:
Or `HTMLView`.

martensievers:
Cool, thanks for your feedback guys. Will merge it as an `HTMLView` so the default HTMLView is a `div` and then others would need to be `HTMLEmbedView` for example if we considered supporting 3rd party HTML in Incito in form of an `iframe` with sandbox etc.

PR #81, opened 2021-03-18
Different approach to rendering the Incito format

mortenbo:
This refactor vastly simplifies the Incito lib and makes parsing + generating HTML elements much faster (around 2-3x times by my comparison). Also, I've consolidated VideoEmbedView and HTMLEmbedView to support iframes. I'm just not sure if we want to support `allow-scripts` (that breaks YouTube et al though).


lib/incito-browser/utils.js
```diff
@@ -2,16 +2,24 @@ export const formatUnit = (unit) => {
...
+    if (typeof value === 'string') {
+        return value.replace(/"/g, '&quot;');
+    }
+
+    return value;
+};
+
 export const isDefinedStr = (value) =>
-    typeof value === 'string' && value.length > 0;
+    value && typeof value === 'string' && value.length > 0;
```
klarstrup: In what case is a `string` longer than 0 falsy?

lib/incito-browser/incito.js
```diff
@@ -1,308 +1,614 @@
...
+            },
+            callback_loading: (el) => {
+                if (
+                    el.tagName.toLowerCase() === 'video' &&
+                    el.getAttribute('data-controls')
+                ) {
+                    el.setAttribute('controls', 'true');
+                }
+            }
+        });
```
klarstrup: Consider using `IntersectionObserver` directly instead of including an 800 line library(that is currently failing lint), something to the effect of

```js
function loadEl(el) {
    el.src = el.dataset.src;
    el.classList.remove('incito--lazy');
    if (
        el.tagName.toLowerCase() === 'video' &&
        el.getAttribute('data-controls')
    ) {
        el.setAttribute('controls', 'true');
    }
}
if ('IntersectionObserver' in window) {
    this.observer = new IntersectionObserver(
        (entries) => {
            entries.forEach((entry) => {
                if (entry.isIntersecting) {
                    loadEl(entry.target);
                    this.observer.unobserve(entry.target);
                }
            });
        },
        {root: null, rootMargin: '400px', threshold: 1.0}
    );
    this.containerEl.querySelector('.incito--lazy').forEach((lazyEl) => {
        this.observer.observe(lazyEl);
    });
} else {
    this.containerEl.querySelector('.incito--lazy').forEach((lazyEl) => {
        loadEl(lazyEl);
    });
}
```

lib/incito-browser/incito.js
```diff
@@ -1,308 +1,614 @@
...
-        window.removeEventListener('resize', this.lazyloader, false);
+            if (view.text_alignment === 'left') {
+                styles['text-align'] = 'left';
+            } else if (view.text_alignment === 'center') {
+                styles['text-align'] = 'center';
+            } else if (view.text_alignment === 'right') {
+                styles['text-align'] = 'right';
+            }
 
-        this.trigger('destroyed');
```
klarstrup, line 134: The Controls of IncitoPublicationKit relies on this `destroyed` event to remove event handlers: https://github.com/shopgun/shopgun-js-sdk/blob/develop/lib/kits/incito-publication/controls.js#L13

lib/incito-browser/incito.js
```diff
@@ -1,308 +1,614 @@
...
+            if (isDefinedStr(view.text_shadow)) {
+                styles['text-shadow'] = view.text_shadow;
+            } else if (textShadow != null) {
+                styles[
+                    'text-shadow'
+                ] = `${textShadow.dx}px ${textShadow.dy}px ${textShadow.radius}px ${textShadow.color}`;
+            }
 
-        document.removeEventListener('scroll', this.lazyloader, true);
-        window.removeEventListener('resize', this.lazyloader, false);
```
klarstrup, line 132: You should still make sure to `.destroy()` the new `this.lazyload` here, so that we don't leak event listeners.

lib/incito-browser/incito.js
```diff
@@ -1,308 +1,614 @@
...
+                styles['text-decoration-line'] = view.text_decoration_line.join(
+                    ' '
+                );
+            }
 
-        return this;
-    }
+            const textShadow = getTextShadow(view);
 
-    destroy() {
```
klarstrup, line 127: `IncitoPublicationKit`'s `Viewer` calls this method for cleanup here https://github.com/shopgun/shopgun-js-sdk/blob/develop/lib/kits/incito-publication/viewer.js

klarstrup (changes requested):
This is a great simplification from the DOM approach, just need some tweaks to not break existing contracts.

lib/incito-browser/utils.js
mortenbo: True - will remove.

lib/incito-browser/incito.js
mortenbo: That's really nice. It just be a slight challenge loading everything immediately for clients (IE11) not supporting IntersectionObserver?

klarstrup: Yeah but that's also what `lazyload.js` does https://github.com/shopgun/shopgun-js-sdk/blob/d80b229d744e7c647a635e2068450fa2d5a4a57b/lib/incito-browser/lazyload.js#L722

Thinking about it, an even better option is for the renderer to detect `IntersectionObserver` support and set `src` instead of `data-src` immediately, then this `} else {` is unnecessary.

mortenbo: @klarstrup That's a great idea. Committed that now.

martensievers:
So we are removing all the separate view files consolidating the view rendering into one single recursive function for all view types which I think is nice and removes some unnecessary complexity but out of curiosity, where do you see the drastic performance increase coming from that you mentioned?

martensievers (approved):
Syntax wise and from the refactor it looks fine to me, don't know much about lazyLoading so I can't comment on that part  ✅

mortenbo:
@martensievers There's multiple reasons for why it's faster:

- String concatenation and use of a single .innerHTML = instead of using the DOM API (document.createElement, etc.)
- Faster font loading (apparently `new FontFace` is slow)
- Improved lazyloading
- Load the first 20 images immediately and not rely on lazyload
- Fixed some weird stuff with the VideoView where setting controls to true for lazyloaded videos resulted in poor performance in Safari

mortenbo:
Also, I removed the CoreKit sessions stuff meaning the first request to fetch Incito's are way faster as we skip a request.

lib/kits/core/request.js
```diff
@@ -1,129 +1,64 @@
...
+    }
 
-                    return `${encodeURIComponent(k)}=${encodeURIComponent(
-                        qs[k]
-                    )}`;
-                })
-                .join('&');
+    if (options.qs) {
+        const url2 = new URL(url);
+        const searchParams = new URLSearchParams(url2.search);
```
klarstrup, line 25: Nice `URL(SearchParams)` API usage replacing all the string operations

lib/kits/core/request.js
```diff
@@ -1,129 +1,64 @@
...
+    const body = options.body;
 
-            if (body) {
-                body = JSON.stringify(body);
-            }
-        }
-
-        if (locale != null) {
-            qs.r_locale = locale;
-        }
```
klarstrup, line 41: Consider removing the accompanying `locale` stuff from the config code, this is the only place it has mattered.

mortenbo: Good idea. Done + appSecret.

lib/incito-browser/incito.js
```diff
@@ -1,308 +1,696 @@
...
+                    rootMargin: '500px'
+                }
             );
+
+            this.el.querySelectorAll('.incito--lazy').forEach((lazyEl) => {
+                this.lazyloader.observe(lazyEl);
+            });
         }
+
+        this.trigger('allRendered');
```
klarstrup: Since this runs totally synchronously now we can ditch this event, knowing that after `Incito.start()` everything will have been rendered.

mortenbo: I just wasn't sure if allRendered was needed by someone?

klarstrup (approved):
Looks great, happy to shed some weight and speed up incito in one go!

Just left some color commentary.

lib/incito-browser/incito.js
mortenbo: I'll remove it.

klarstrup: Not to my knowledge, but it's very hard for us to know. Easy fix if anyone was though, just move `allRendered` event handler code after `viewer.start()`

Introduced here btw https://github.com/shopgun/incito-browser/commit/ff28c61383b829f6a8affc5befbd7d19e984f21a

PR #83, opened 2021-04-04
Simplify experiments for others to use

mortenbo:


lib/kits/incito-publication/controls.js
```diff
@@ -16,14 +15,14 @@ export default class Controls {
...
-        const winHeight = window.innerHeight;
-        const rect = this.viewer.el.getBoundingClientRect();
-        const progress = Math.min(
-            100,
-            Math.round((Math.abs(rect.top - winHeight) / rect.height) * 100)
+        const progress = Math.round(
+            (window.pageYOffset /
+                (document.body.scrollHeight - window.innerHeight)) *
+                100
         );
```
klarstrup, line 22: Beware that this is functionally different, and will give inaccurate %s if the page has stuff above or below the incito. (And never reach 100% I think?)

klarstrup (approved):
LGTM - one note

lib/kits/incito-publication/controls.js
mortenbo: @klarstrup That is true but I just opted out of it so the scrollbar matches the percentage.

klarstrup: As long as it is intentional :]

mortenbo: It is and it's just a demo page so they can always alter it so their needs.

klarstrup: But this is not demo page specific :/ it's used by `IncitoPublicationKit` if a `.sgn-incito__progress` element is present within the viewer target - Fakta and Elgiganten and probably several others are using this feature on their own websites right now.

mortenbo: Right, you are totally right. I still think it's best like this tbh. Do you feel it's better the other way?

PR #84, opened 2021-04-14
Add appKey config requirement in CoreKit.request

klarstrup:
As discussed at https://github.com/shopgun/TjekTeams/issues/57#issuecomment-818677971

This error is added at the same place the old `'Invalid API key.\nCould not find app matching your api key.'` error would have happened, when this was a backend requirement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ When installed you can require and configure the SDK:
 SGN = require('sgn-sdk');
 
 SGN.config.set({
-    appKey: 'YOUR_APP_KEY'
+    apiKey: 'YOUR_API_KEY'
 });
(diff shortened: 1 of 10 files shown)


mortenbo (approved):
Looks good. Consider adding `apiKey` as an option in the SDK besides `appKey` to migrate to apiKey naming. They do the same thing but then it's 1:1 between the developer portal and this SDK.

mortenbo (approved):
👏

klarstrup:
Obs. I've left `data-app-key` in a couple of old examples targeting SDK versions 2.2.5 and 3.3.0.

PR #86, opened 2021-06-22
get first frame for poster

t-lansing:


t-lansing:
hi, @klarstrup. the idea here is to get first video frame and display it as a video poster, so that we always show an image from the video in the Incito - rather than a grey box when the video is loading. 

I haven't tested it, since I am not familiar with the local dev setup. maybe you can give me a brief intro to that. thanks.

klarstrup:
Hey! Easiest way to run this locally is 

```bash
git clone git@github.com:shopgun/shopgun-js-sdk.git
cd shopgun-js-sdk
npm install
npm run dev
```

and then opening `shopgun-js-sdk/examples/incito-publication.html` in a browser

lib/incito-browser/incito.js
```diff
@@ -292,7 +291,7 @@ export default class Incito extends MicroEvent {
             attrs.muted = '';
             attrs.playsinline = '';
             attrs.preload = 'none';
-            attrs.poster = 'noposter';
+            attrs.poster = view.src + '#t=0.1';
```
klarstrup: I don't think this will work as expected, but I didn't test it. Wouldn't this attribute need to point to an image?

mortenbo: I don’t think this is the way either.

t-lansing:
@klarstrup @mortenbo by removing the poster attribute entirely, the first-frame poster would get created automatically by the browser:

poster
A URL for an image to be shown while the video is downloading. **If this attribute isn't specified, nothing is displayed until the first frame is available, then the first frame is shown as the poster frame.**

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video

This does not improve our overall video situation much, as the poster seems to be visible mainly when there is no autoplay and the first frame arrives late anyway on a slow connection. 

I can look further into the 'rocket' approach - i.e. showing videos in multiple resolutions, starting by the low-res one first, and ditching that once the high res is available.

mortenbo:
@t-lansing Yeah, I think the rocket approach is the only real solution tbh.

klarstrup:
Can't our CDN generate a first frame?

PR #87, opened 2021-08-29
feat: IncitoEmbedView view + global styles

mortenbo:
This view can recursively fetch more Incito's. This allows you to split up an Incito into smaller chunks that can be lazyloaded and thereby reduce number of bytes users have to download + allow for an instant loading experience.

Also: introduce global theme styles to reduce payload size even further.


lib/incito-browser/incito.js
```diff
@@ -368,8 +380,8 @@ export default class Incito extends MicroEvent {
             } else {
                 attrs.src = view.src;
             }
-        } else if (view.view_name === 'IncitoEmbedVideo') {
-            classNames.push('incito__incito-embed-video');
+        } else if (view.view_name === 'IncitoEmbedView') {
```
klarstrup, line 383: Is this removal of `IncitoEmbedVideo` views cleared with Incito team?

lib/incito-browser/incito.js
```diff
@@ -48,6 +48,14 @@ export default class Incito extends MicroEvent {
             this.el.style.color = theme.text_color;
         }
 
+        if (isDefinedStr(theme.style)) {
+            this.styleEl = document.createElement('style');
+
+            this.styleEl.innerText = theme.style;
+
+            document.head.appendChild(this.styleEl);
+        }
```
klarstrup, line 57: Incito team needs to be super diligent not to mess with ETA & client styles, of course

lib/incito-browser/incito.js
mortenbo: @klarstrup That was a typo by me (I wrote `IncitoEmbedVideo` when I meant `IncitoEmbedView`). We still support `VideoEmbedView` and `HTMLEmbedView`.

lib/incito-browser/incito.js
mortenbo: @klarstrup Indeed. I wish we could scope the CSS. There was a proposal for `scoped` style attribute but it's deprecated. Do you have a better approach? This seemed like the best one to me.

mortenbo:
@martensievers @simonmarqvard @t-lansing  OK to merge or do you have some feedback?

martensievers:
@mortenbo I think it looks good. The option for global styles could be useful and setting up the structure for sending Incito's in page chunks seems like a good direction to go in 👍🏼

lib/incito-browser/incito.js
klarstrup: We could scope the CSS like this: 

1. put the incito ID as a class on the incito mount point i.e. `<div class="sgn__incito incito__moBP88ob"></div>`
2. have bumblebee prefix all the styles with `.incito__moBP88ob <actual selector goes here>`(i.e. `.incito__moBP88ob h1 {font-size: xxx-large;})`
    3. or modify `theme.style` here to add `.incito__moBP88ob` to all rules (will require a lot of parsing and string munging)


(sorry about the late reply)

mortenbo: That could work but if we have a property for global styles and we want to prepend an Incito ID in front, wouldn't we have to parse all the CSS and do string munging, as you say?

PR #89, opened 2021-09-11
First stab at rename

mortenbo:
We need to address:

- S3 upload
- How to avoid singleton (@klarstrup you had some ideas here, I know)
- Deprecation notices / readme updates
- SDK upgrade guide


klarstrup (commented):
What is the value add here? There should be some tangible benefits to integrators if we are to expect them to rename a bunch of things

I think we should examine the needs of integrators and support exactly those features in `@tjek/sdk`,

mortenbo:
@klarstrup While I agree, my desire is to move away from the name ShopGun as it's confusing for integrators. We're called Tjek and having ShopGun as an extra brand that isn't utilized anymore adds to that confusion, IMHO. We can also keep the `sgn` shorthand and alias it with `Tjek` and then rename this repo. That could also work.

You already have @tjek/sdk defined in the manifest so I thought we might as well commit to it.

Do you have another idea of how we may achieve my goal of retiring ShopGun and using Tjek?

PR #91, opened 2021-09-13
Make event-tracker-pool instantiation lazy and Tracker module pure

### LINKED ISSUE #90: [EventsKit] Make event-tracker-pool instantiation lazy
Currently `event-tracker-pool` is set on `localStorage` on all sites with `shopgun-sdk` included in any way, regardless of whether an EventTracker is even active on the page. The code also runs serverside on any node server that `require('shopgun-sdk');` (but doesn't do anything except cost a few cycles to throw a couple of caught errors).

klarstrup:


klarstrup:
Consider writing some E2E puppeteer tests using request interception

lib/kits/events/tracker.js
```diff
@@ -20,19 +20,28 @@ const createTrackerClient = () => {
...
+    return Array.isArray(data)
+        ? data.filter((evt) => typeof evt._i === 'string')
+        : [];
 }
 
-let pool = getPool();
+const unloadHandler = () =>
+    clientLocalStorage.set('event-tracker-pool', pool.concat(getPool()));
 
+let pool;
```
klarstrup, line 33: The `pool` is however still a singleton. One issue I can think of there is if an integrator were to be using multiple `eventsTrackUrl`s, which doesn't seem very likely but stranger things have happened...

The extreme version of this PR would maybe have each Tracker have its own `pool` instance property

mortenbo (approved):
Good job. Did you run lint and prettier on it? Then I don't have to nit 😎

PR #93, opened 2021-09-17
Traverse the DOM for data-link

mortenbo:
The issue was you could have a parent view with a link with a TextView inside. If you happened to click the TextView, the click handler wouldn't trigger as that target view didn't have the data-link attribute. Using closest(), we'll traverse upwards until we hit a data-link.

### DIFF
--- a/lib/incito-browser/incito.js
+++ b/lib/incito-browser/incito.js
@@ -12,6 +12,7 @@ import {
     getTextShadow,
     loadFonts
 } from './utils.js';
+import {closest} from '../util.js';
 
 export default class Incito extends MicroEvent {
     constructor(containerEl, {incito = {}}) {
@@ -76,9 +77,8 @@ export default class Incito extends MicroEvent {
 
     start() {
         this.el.addEventListener('click', (e) => {
-            const link = e.target.closest
-                ? e.target.closest('[data-link]').getAttribute('data-link')
-                : e.target.getAttribute('data-link');
+            const el = closest(e.target, '[data-link]');
+            const link = el ? el.dataset.link : null;
 
             if (isDefinedStr(link)) {
                 window.open(link, '_blank');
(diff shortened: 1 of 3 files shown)


martensievers (approved):
Lgtm 👍🏼

lib/incito-browser/incito.js
```diff
@@ -76,7 +76,9 @@ export default class Incito extends MicroEvent {
 
     start() {
         this.el.addEventListener('click', (e) => {
-            const link = e.target.getAttribute('data-link');
+            const link = e.target.closest
+                ? e.target.closest('[data-link]').getAttribute('data-link')
+                : e.target.getAttribute('data-link');
```
klarstrup: Replying here instead of on the DM ;)

`@babel/preset` & `core-js` does not include polyfills for specific DOM features like this, I would propose a little something like this: (It's the MDN recommended polyfill changed to be a ponyfill instead, so we don't patch prototypes on client websites)

```js
const matches =
    Element.prototype.matches ||
    Element.prototype.msMatchesSelector ||
    Element.prototype.webkitMatchesSelector;

function closest(el, s) {
    do {
        if (matches.call(el, s)) return el;

        el = el.parentElement || el.parentNode;
    } while (el !== null && el.nodeType === 1);

    return null;
}
```

used like

```js
const link = closest(e.target, '[data-link]').dataset.link;
```

But do be mindful that `closest` can return null and crash this example and yours.

klarstrup: Additionally it might be worth checking that the `[data-link]` element is inside the incito, in case integrators use data-link on their own site for whatever reason: 
```js
const link = closest(e.target, '.incito__view [data-link]').dataset.link;
```

klarstrup (changes requested):

lib/incito-browser/incito.js
mortenbo: @klarstrup Lovely feedback. Much appreciated. I should have addressed it now.

klarstrup (approved):
Excellent

PR #95, opened 2021-10-25
refactor: change hyphens behaviour to manual

martensievers:
Adjust the `incito-browser` hyphens settings from `auto` to `manual` to prevent forced line-breaks in offer details.


lib/incito-browser/incito.styl
```diff
@@ -38,7 +38,7 @@
 
 .incito__text-view
     margin 0
-    hyphens auto
+    hyphens manual
```
mortenbo: Shouldn't it just be omitted?

martensievers: I have removed it completely now 👍🏼

martensievers:
@klarstrup can we merge this and get a new minor release out?

klarstrup:
Keep in mind that this will affect incitos of the past as well

Did you consider letting it be controlled through styles so we can change the behavior just going forward?

martensievers:
> Keep in mind that this will affect incitos of the past as well
> 
> Did you consider letting it be controlled through styles so we can change the behavior just going forward?

I'm not entirely sure what you mean by "setting it but through styles"?

klarstrup:
like you set any other incito element style attribuet

martensievers:
> like you set any other incito element style attribuet

I think we are seeing that in all cases it's better without the hyphens, so it seems like making changes to the incito format to control the styling on an individual view is not necessary right now.

martensievers:
Can we make a decision on this to get it out please? This will really improve the visual experience of Incito a lot and it's been ready for deploy for several weeks now.

Adding this to the styles in incito would require an update to the incito format correct @mortenbo? I don't know if that effort is in any relation to the benefits of having this setting controllable on a view level tbh.

mortenbo:
@martensievers Jonas just means you can set hyphens manually for TextView's but let's just change the default so that hyphens is not enabled by default and you have to explicitly set it (normal text behavior).

@klarstrup If you agree, could you make a new version?

martensievers:
> @martensievers Jonas just means you can set hyphens manually for TextView's but let's just change the default so that hyphens is not enabled by default and you have to explicitly set it (normal text behavior).
> 
> @klarstrup If you agree, could you make a new version?

@mortenbo but don't I have to define that in the incito format first and then add a check in the browser to check for the existence etc.?

martensievers:
Agree, let's disable it by default and then we can set it explicitly as text style if needed.

klarstrup (approved):

PR #100, opened 2021-12-09
Feature/generic template builder

jrlarano:
Paged Publication supported features:
`/examples/paged-publication-template`
- [x] View publication
- [x] Offer list
- [x] Page list
- [x] Search offer
- [x] Shopping list
- [x] Print shopping list
- [x] Publication download
- [x] Store list

List Publication supported features:
`/examples/list-publications-template`
- [x] View publication list
- [x] Display paged publication on click


lib/config.js
```diff
@@ -30,7 +30,8 @@ Config.prototype.keys = [
     'authToken',
     'eventTracker',
     'coreUrl',
-    'eventsTrackUrl'
+    'eventsTrackUrl',
+    'businessId'
```
mortenbo: Should this go away?

lib/kits/core-ui/components/helpers/component.js
```diff
@@ -0,0 +1,51 @@
+const componentHelper = {
+    addBlockerListenerTo: (container, destroyCallback) => {
+        const headNav = document.querySelector('.sgn__nav');
+        const blocker = document.createElement('div');
+        blocker.className = 'sgn-blocker';
```
mortenbo: Some spacing between variable declarations and assignments perhaps.

lib/kits/core-ui/components/helpers/component.js
```diff
@@ -0,0 +1,51 @@
...
+                componentHelper.destroyBlocker(container);
+
+                if (typeof destroyCallback === 'function') {
+                    destroyCallback(e);
+                }
+            },
+            false
+        );
+    },
+    destroyBlocker: (container) => {
```
mortenbo: What is "blocker"?

lib/kits/core-ui/components/paged-publication/main-container.js
```diff
@@ -0,0 +1,174 @@
...
+const defaultShoppingCountTemplate = `\
+{{#shoppingCount}}
+    <div class="sgn__offer-shopping-counter">
+        <span>{{shoppingCount}}</span>
+    </div>
+{{/shoppingCount}}\
+`;
+
+const mainContainer = ({template, shoppingCounterTemplate, el, scriptEls}) => {
+    const _template = template?.innerHTML || defaultTemplate;
```
mortenbo: Could you elaborate on why you use _ in front of variables?

lib/kits/core-ui/components/paged-publication/main-container.js
```diff
@@ -0,0 +1,174 @@
...
+                .forEach((name) => {
+                    const element = _el.querySelector(`.${name}`);
+                    if (element) {
+                        element.parentNode.removeChild(element);
+                    }
+                });
+        }
+    };
+
+    const renderShoppingCount = () => {
```
mortenbo: In general, we should strive towards called it `shopping list`.

lib/kits/core-ui/components/paged-publication/menu-popup.js
```diff
@@ -0,0 +1,163 @@
...
+    const _translations = {
+        localeCode: translate('locale_code'),
+        untilLabel: translate('paged_publication_viewer_until_label'),
+        pagesButton: translate('paged_publication_viewer_pages_button'),
+        offersButton: translate('paged_publication_viewer_offers_button'),
+        downloadButton: translate('paged_publication_viewer_download_button')
+    };
+
+    const render = async () => {
+        console.log('sgnData', _sgnData);
```
mortenbo: Remove debug info.

lib/kits/core-ui/components/paged-publication/offer-list.js
```diff
@@ -0,0 +1,200 @@
...
+
+const defaultTemplate = `\
+    <div class="sgn-offers-content">
+        <div class="sgn-offers-search-container">
+            <input type="text" id="sgn-offers-search-text" class="sgn-offers-search-text" name="sgn-offers-search-text" placeholder="{{translations.searchText}}...">
+        </div>
+        <ol class="sgn-offers-ol">
+            {{#offers}}
+            <li class="sgn-offers-li">
+                <div class="sgn-offers-li-flex-container" data-page-number="{{catalog_page}}">
```
mortenbo: Make sure we use publication (or paged publication or Incito publication) instead of catalog.

lib/kits/core-ui/components/paged-publication/offer-list.js
```diff
@@ -0,0 +1,200 @@
...
+                    <div class="sgn-offers-li-div-img">
+                        <img src="{{images.view}}" alt="{{heading}}">
+                    </div>
+                    <div class="sgn-offers-li-div-text">
```
mortenbo: I would advise against using tag names as class names (e.g. div).

lib/sgn-sdk.js
```diff
@@ -71,8 +73,19 @@ if (isBrowser()) {
...
             config.eventTracker = new SGN.EventsKit.Tracker({trackId});
         }
+        if (businessId) {
+            config.businessId = businessId;
+        }
 
         SGN.config.set(config);
+
+        if (component === 'view-publication') {
+            SGN.CoreUIKit.PagedPublication(scriptEl).render();
```
mortenbo: We need to pave the way for Incito too. So perhaps component should be view-paged-publication OR the SDK handles which one to use automatically.

lib/storage/client-local.js
```diff
@@ -22,5 +22,8 @@ export function get(key) {
 export function set(key, value) {
     try {
         storage[`${prefixKey}${key}`] = JSON.stringify(value);
+
+        const event = new Event('storageUpdate');
```
mortenbo: Why this?

locales/da_DK.js
```diff
@@ -0,0 +1,14 @@
+export default {
+    locale_code: 'da-DK',
```
mortenbo: Why have locale code inside the locale and why da-DK instead of da_DK? We use ll_CC everywhere.

locales/en_US.js
```diff
@@ -0,0 +1,14 @@
+export default {
+    locale_code: 'en-US',
+    paged_publication_viewer_shopping_list_label: 'Shopping List',
+    paged_publication_viewer_clear_list_button: 'Clear list',
```
mortenbo: Let's use `shopping list`.

locales/da_DK.js
```diff
@@ -0,0 +1,14 @@
...
+    paged_publication_viewer_shopping_list_label: 'Indkøbsliste',
+    paged_publication_viewer_clear_list_button: 'Ryd liste',
```
mortenbo: liste -> indkøbsliste

mortenbo (changes requested):
Overall, amazing job! Have some feedback.

lib/kits/core-ui/components/helpers/component.js
jrlarano: Blocker is the container outside the modal that has click listener to close the modal. Maybe destroyModal is a better name 😄

lib/kits/core-ui/components/paged-publication/main-container.js
jrlarano: This is only for the distinction between local and global variables especially when I was converting it from classes to functions. We can just remove it if you want to? :)

lib/storage/client-local.js
jrlarano: The only purpose of this is to trigger an event that will update the shopping list counter.

locales/da_DK.js
jrlarano: This is to avoid replacing - to _ everytime we use built in functions eg: date formatters uses `da-DK` format, but we can change this to _ for uniformity 😄

lib/storage/client-local.js
mortenbo: OK. Since this is going to run on 3rd party websites we need to prefix these things. Could be `tjek_shopping_list_update`

lib/kits/core-ui/components/paged-publication/main-container.js
mortenbo: I'm thinking: 1) We shouldn't have global variables (should be within SGN) 2) _ is not pretty so if we can avoid it it would be nice 😅

jrlarano: Sure, I'll refactor it to remove those _ :D

examples/paged-publication-template.html
```diff
@@ -0,0 +1,62 @@
...
+        -->
+
+        <script
+            src="../../dist/shopgun-sdk/sgn-sdk.js?t=3"
+            id="sgn-sdk"
+            data-api-key="GL0wU8"
+            data-track-id="2N2Gib"
+            data-business-id="11deC"
+            data-publication-id-query-param="id"
+            data-component="view-paged-publication"
```
mortenbo: Nitpick: paged-publication-viewer

lib/kits/core-ui/components/list-publications/main-container.js
```diff
@@ -0,0 +1,94 @@
...
+        untilLabel: translate('paged_publication_viewer_until_label')
+    };
+
+    const fetchPublications = () => {
+        return new Promise((resolve) => {
+            request(
+                {
+                    url: `/v2/catalogs`,
+                    qs: {
+                        dealer_id: configs.dealerId,
```
mortenbo: Where does this come from? Shouldn't it be businessId?

lib/kits/core-ui/paged-publication.js
```diff
@@ -0,0 +1,210 @@
...
+                PagedPublicationComponents.MenuPopup({
+                    configs: options,
+                    sgnData,
+                    sgnViewer,
+                    templates: customTemplates
+                }).render();
+            });
+        },
+
+        animateshoppingListCounter: () => {
```
mortenbo: animateS*

lib/kits/core-ui/paged-publication.js
```diff
@@ -0,0 +1,210 @@
...
+                            bootstrapper.applyHotspots(sgnViewer, hotspots);
+                        }
+                    });
+                } else {
+                    console.error(err);
+                }
+            });
+        },
+
+        fetchLatestCatId: () => {
```
mortenbo: Can you make sure we use the word publication instead of cat/catalog?

locales/da_DK.js
```diff
@@ -0,0 +1,14 @@
+export default {
+    locale_code: 'da_DK',
+    paged_publication_viewer_shopping_list_label: 'Indkøbsliste',
+    paged_publication_viewer_shopping_list_clear_button: 'Ryd indkøbsliste',
+    paged_publication_viewer_delete_crossed_out_button:
+        'Slet overstregede elementer',
```
mortenbo: elementer => varer

mortenbo (changes requested):
Looks good. Only smaller nitpicks. Do you have an updated README.md too?

jrlarano:
> 

Nothing yet. I'll try to update the README.md tomorrow :)

lib/kits/core-ui/components/list-publications/main-container.js
jrlarano: It came from here: https://github.com/shopgun/shopgun-js-sdk/blob/e06e072ac2dadb93b5ecf6e7af1fc50a0b2af300/lib/kits/core-ui/list-publications.js#L28

mortenbo: OK, let's update it there too then.

mortenbo (approved):
Nice!

lib/kits/core-ui/components/helpers/component.js
```diff
@@ -0,0 +1,58 @@
+const componentHelper = {
```
klarstrup: Please `export`(and `import`) each of these helpers separately, having a big object like this is not conducive to dead code elimination(and eventually splitting up the SDK)

lib/kits/core-ui/components/paged-publication/publication-download.js
```diff
@@ -0,0 +1,32 @@
...
+                        resolve(publication.pdf_url);
+                    } else {
+                        reject(err);
+                    }
+                }
+            );
+        }).then((url) => {
+            location.href = url;
+        });
+    };
```
klarstrup: Use async/await when possible:
```js
const downloadPdf = async () => {
    location.href = await request({url: `/v2/catalogs/${configs.id}/download`});
};
```

lib/kits/core-ui/components/list-publications/main-container.js
```diff
@@ -0,0 +1,94 @@
...
+                    }
+                },
+                (err, publications) => {
+                    if (!err) {
+                        resolve(transformPublications(publications));
+                    }
+                }
+            );
+        });
+    };
```
klarstrup: Use async/await when possible

lib/kits/core-ui/components/paged-publication/offer-list.js
```diff
@@ -0,0 +1,204 @@
...
+                    }
+                },
+                (err, offers) => {
+                    if (!err) {
+                        resolve(transformOffers(offers));
+                    }
+                }
+            );
+        });
+    };
```
klarstrup: Use async/await when possible

lib/kits/core-ui/components/paged-publication/shopping-list.js
```diff
@@ -0,0 +1,285 @@
...
+
+        shareBtn?.addEventListener('click', () => {
+            if (navigator.share) {
+                navigator
+                    .share(shareData)
+                    .then(() => {
+                        console.log('Shared data:', shareData);
+                    })
+                    .catch(console.error);
+            }
```
klarstrup, line 274: Firefox and Safari like it if you ask `navigator.canShare(shareData)` first, but many other browsers don't have it, so you need to check if it exists and ask first if it does but not if it doesn't

lib/kits/core-ui/list-publications.js
```diff
@@ -0,0 +1,73 @@
+import * as ListPublicationsComponents from './components/list-publications';
```
klarstrup: just import MainContainer if that's all you use

lib/kits/core-ui/paged-publication.js
```diff
@@ -0,0 +1,206 @@
+import * as PagedPublicationKit from '../paged-publication';
```
klarstrup: Just import Bootstrapper if that's all you use

lib/kits/core-ui/paged-publication.js
```diff
@@ -0,0 +1,206 @@
...
+import * as PagedPublicationComponents from './components/paged-publication';
```
klarstrup: Directly import
`import { MainContainer, ShoppingList, MenuPopup } from './components/paged-publication';`

lib/kits/core-ui/paged-publication.js
```diff
@@ -0,0 +1,206 @@
...
+            if (!options.id) {
+                options.id =
+                    await pagedPublicationObj.fetchLatestPublicationId();
+            }
+        },
+
+        start: () => {
+            const bootstrapper = new PagedPublicationKit.Bootstrapper(options);
+
+            bootstrapper.fetch((err, data) => {
```
klarstrup: You could skip `bootstrapper.fetch` here since you already can have the relevant publication in hand with `.fetchLatestPublication`. Just fetch `catalogs/${publication.id}/pages` too and then you can call into `createViewer` like this:

```js
sgnViewer = bootstrapper.createViewer({
    details: publication,
    pages: await request({url: `catalogs/${publication.id}/pages`})
});
```

Saves an unnecessary `catalogs/${publication.id}` request that `bootstrapper.fetch` wastes its time doing here

lib/sgn-sdk.js
```diff
@@ -73,6 +74,14 @@ if (isBrowser()) {
         }
 
         SGN.config.set(config);
+
+        if (component === 'paged-publication-viewer') {
+            SGN.CoreUIKit.PagedPublication(scriptEl).render();
+        }
+
+        if (component === 'list-paged-publications') {
+            SGN.CoreUIKit.ListPublications(scriptEl).render();
```
klarstrup: I would prefer using `CoreUIKit.ListPublications` here(and `CoreUIKit.PagedPublication` above), we try not to use the singleton within the SDK as it will make it difficult to component-ize the SDK when we get around to it.

lib/storage/client-local.js
```diff
@@ -22,5 +22,9 @@ export function get(key) {
 export function set(key, value) {
     try {
         storage[`${prefixKey}${key}`] = JSON.stringify(value);
+
+        const event = new Event('tjek_shopping_list_update');
+
+        window.dispatchEvent(event);
```
klarstrup: It's not okay for everything in the SDK that calls `clientStore.set()` to dispatch this event

klarstrup (commented):
Super cool stuff this is going to make adoption a breeze

I didn't want to be too obnoxious with the comments so i stopped writing them but our `request` helper supports `async`/`await`, i put an example in one of the comments. 

Additionally try not to over-import code, we need to be mindful of file size. Currently the build is failing because the files grew by quite a bit: https://github.com/shopgun/shopgun-js-sdk/runs/4947810595?check_suite_focus=true

To make tests pass so we can get this merged you will need to update the `bundlesize` `maxSize`s listed in `package.json` accordingly.

lib/kits/core-ui/paged-publication.js
```diff
@@ -0,0 +1,206 @@
...
+        const mainContainerEl = document.querySelector(scriptEls.mainContainer);
+
+        PagedPublicationComponents.MainContainer({
+            template: customTemplates.mainContainer,
+            shoppingListCounterTemplate: customTemplates.shoppingListCounter,
+            el: mainContainerEl,
+            scriptEls
+        }).render();
+    };
+    renderMainContainer();
```
klarstrup: Why not just

```js
        const mainContainerEl = document.querySelector(scriptEls.mainContainer);

        PagedPublicationComponents.MainContainer({
            template: customTemplates.mainContainer,
            shoppingListCounterTemplate: customTemplates.shoppingListCounter,
            el: mainContainerEl,
            scriptEls
        }).render();
```

lib/kits/core-ui/paged-publication.js
```diff
@@ -0,0 +1,206 @@
...
+
+    const pagedPublicationObj = {
+        render: async () => {
+            if (Object.keys(options).length === 0) {
```
klarstrup: I think almost every method on this object can be a free function instead, from what I can see only `render` is used by `pagedPublication` callers - we don't need to expose all these other functions to consumers(or risk them not being code eliminated if unused).

lib/kits/core-ui/components/paged-publication/main-container.js
```diff
@@ -0,0 +1,173 @@
...
+                    }
+                });
+        }
+    };
+
+    const renderShoppingListCounter = () => {
+        const template =
+            shoppingListCounterTemplate?.innerHTML ||
+            defaultShoppingListCounterTemplate;
+        const clientStorage = clientLocalStorage.get(
```
klarstrup: Please use a descriptive name here

lib/kits/core-ui/components/list-publications/main-container.js
```diff
@@ -0,0 +1,94 @@
...
+                </div>
+            </div>
+        </li>
+        {{/publications}}
+    </u;>
+</div>
+<div id="sgn-paged-publication-container"></div\
+`;
+
+const mainContainer = ({configs, template, el}) => {
```
klarstrup: This is used as `MainContainer` everywhere so it's probably best to make that naming consistent

lib/kits/core-ui/components/helpers/translate.js
```diff
@@ -0,0 +1,38 @@
+import * as locales from '../../../../../locales';
+
+const getTranslationOverride = (dataset = {}) => {
+    const keys = Object.keys(dataset);
+    const translationsKeyVal = {};
+    keys.filter((key) => key.includes('translationKeys-')).map((key) => {
```
klarstrup: use `forEach` instead of `map` if you're only looping for side-effects and not using the new array that `map` returns.

lib/kits/core-ui/components/helpers/translate.js
```diff
@@ -0,0 +1,38 @@
+import * as locales from '../../../../../locales';
+
+const getTranslationOverride = (dataset = {}) => {
```
klarstrup: I would've written this using `Object.entries` and `Array#reduce`:

```js
const getTranslationOverride = (dataset = {}) =>
    Object.entries(dataset).reduce((translationsKeyVal, [key, value]) => {
        if (key.includes('translationKeys-')) {
            const newKey = key
                .replace('translationKeys-', '')
                .replace(/(?:^\w|[A-Z]|\b\w)/g, (word, index) =>
                    index === 0 ? word.toLowerCase() : `_${word.toLowerCase()}`
                );

            translationsKeyVal[newKey] = value;
        }

        return translationsKeyVal;
    }, {});
```

lib/kits/core-ui/paged-publication.js
```diff
@@ -0,0 +1,206 @@
...
+                            order_by: '-publication_date',
+                            limit: 1
+                        }
+                    },
+                    (err, publications) => {
+                        if (!err) {
+                            const config = {
+                                publicationId: publications[0].id
+                            };
+                            SGN.config.set(config);
```
klarstrup: I think this is polluting the global config, plus I can't see where you're using this `publicationId` anywhere?

lib/kits/core-ui/paged-publication.js
```diff
@@ -0,0 +1,206 @@
...
+                eventTracker: SGN.config.get('eventTracker'),
+                pageId:
+                    opts.pageId ||
+                    `page${SGN.util.getQueryParam(scriptEls.pageIdParam) || 1}`,
+                id:
+                    opts.id ||
+                    SGN.util.getQueryParam(scriptEls.publicationIdParam) ||
+                    scriptEls.publicationId ||
+                    '',
+                businessId: opts.businessId || scriptEls.businessId
```
klarstrup: You could probably extract all these properties as free variables in the `pagedPublication` closure

lib/kits/core-ui/paged-publication.js
jrlarano: Do you think it will be fine if we maintain this as it is for now? `.fetchLatestPublication` is only a fallback when there's no available publicationId being passed as parameter (url param, or param from publications list view).

lib/kits/core-ui/components/list-publications/main-container.js
jrlarano: All function names are in camel case on declaration only, but they are capitalized when exported. Should we make all function names in capital?

klarstrup: I don't feel so strongly about that, as long as the same function doesn't have different names in different places.

lib/kits/core-ui/components/helpers/component.js
```diff
@@ -0,0 +1,114 @@
...
+
+export {
+    insertAfter,
+    insertInside,
+    destroyModal,
+    addBlockerListenerTo,
+    formatPrice,
+    formatDate,
+    translate
+};
```
klarstrup: Please split these into individual exports e.g. `export const insertAfter =` etc.

lib/kits/core-ui/paged-publication.js
```diff
@@ -0,0 +1,224 @@
...
+    const setOptions = async (opts = {}) => {
+        options = {
+            el: document.querySelector('.sgn__pp'),
+            eventTracker: SGN.config.get('eventTracker'),
+            pageId:
+                opts.pageId ||
+                `page${SGN.util.getQueryParam(scriptEls.pageIdParam) || 1}`,
+            id:
+                opts.id ||
+                SGN.util.getQueryParam(scriptEls.publicationIdParam) ||
```
klarstrup: Please import these two SGN utilities directly from `../../util` instead, it is best we avoid using this singleton internally.

PR #102, opened 2022-02-04
Feature/generic template enhancements

jrlarano:


examples/paged-publication-template.html
```diff
@@ -53,8 +53,14 @@
             data-publication-id-query-param="id"
             data-component-theme = "light"
             data-component-paged-publication-remove-components = "classnames_to_be_removed_separated_by_comma"
+            data-component-paged-publication-disable-shopping-list = "false"
+            data-component-paged-publication-disable-close = "false"
```
mortenbo: I'd replace ` = ` with `=` (no spaces)

lib/kits/core-ui/components/list-publications/main-container.js
```diff
@@ -50,17 +51,32 @@ const MainContainer = ({configs, template, el}) => {
...
+                            publication[key] === value && prev,
+                        {}
+                    )
+                )
+                ?.map((publication) => {
+                    const date_from = formatDate(
+                        publication?.run_from,
+                        localeCode
+                    );
+                    const date_till = formatDate(
```
mortenbo: Should this happen here?

mortenbo: Btw, we use camelCase and not snake_case.

lib/kits/core-ui/components/paged-publication/main-container.js
```diff
@@ -154,6 +153,20 @@ const MainContainer = ({
             e.stopPropagation();
             sgnPp?.parentNode?.removeChild(sgnPp);
         });
+
+        sgnPp.addEventListener(
```
mortenbo, line 157: It's important when doing escape logic that the current view always has focus() and tabindex="-1". So when you close the menu, it's important you focus() the publication viewer both so that arrows work for navigation and that escape works.

mortenbo (approved):

lib/kits/core-ui/paged-publication.js
```diff
@@ -168,10 +188,15 @@ const PagedPublication = (scriptEl, mainContainer = '') => {
...
-            scriptEls.offerClickBehavior === 'open-offer-link' &&
+            scriptEls.offerClickBehavior === 'open_webshop_link_in_tab' &&
             hotspot?.webshop
         ) {
             window.open(hotspot.webshop);
+        } else if (
+            scriptEls.offerClickBehavior === 'redirect_to_webshop_link' &&
+            hotspot?.webshop
+        ) {
+            window.open(hotspot.webshop, '_self');
```
klarstrup: How is this different from the ordinary `location.href = hotspot.webshop;`?

klarstrup (approved):

lib/kits/core-ui/paged-publication.js
jrlarano: Nothing really. But I just read that location.href might be better so I'll change it to that :D

PR #108, opened 2022-03-10
[Template] Fix Paged viewer overflow

jrlarano:


klarstrup (changes requested):
Can we accomplish this without messing with CSS outside of our component? :/

klarstrup (changes requested):
This still applies global styling to a website that is not ours, that's gonna bite us in the hind eventually

mortenbo:
@klarstrup Isn't this the only way to disable scroll on the site?

klarstrup:
> @klarstrup Isn't this the only way to disable scroll on the site?

if the goal is to disable scrolling, we should capture scroll inputs on our component and block them as appropriate

jrlarano:
> > @klarstrup Isn't this the only way to disable scroll on the site?
> 
> if the goal is to disable scrolling, we should capture scroll inputs on our component and block them as appropriate

The goal is not only the for the scroll inputs but to also hide the scrollbar when publication viewer is open. Is there any other way apart from this?

klarstrup:
> > > @klarstrup Isn't this the only way to disable scroll on the site?
> > 
> > 
> > if the goal is to disable scrolling, we should capture scroll inputs on our component and block them as appropriate
> 
> The goal is not only the for the scroll inputs but to also hide the scrollbar when publication viewer is open. Is there any other way apart from this?

yeah stay within 100vh, you can also use position: fixed; to position the container edge-to-edge in the viewport

PR #111, opened 2022-03-30
Add option to display query params

jrlarano:
- Add option to disable opening of publication viewer
- Dispatch event upon clicking publication from list
- Add option to display publication id and page number as query param
- Add option to display publication id and page number as URL hash fragment
    `data-component-publication-display-url-params="hash || query"`

eg: 
`http://localhost:3000/examples/list-publications-template`
`http://localhost:3000/examples/paged-publication-template`


mortenbo (approved):
Cool! If @klarstrup agrees with this approach too, I'm on board.

klarstrup:
I think we should put these details in the `URL#hash` instead, that can't collide with client solutions in the same way. 

I know this proposal is opt-in so nothing breaking off the bat, but if integrators don't test it properly they might find themselves with a semi-broken solution. Imagine a JSP application that errors on unexpected query parameters, then we add some of ours client side and only when the user tries to refresh or shares the link will an error happen.

jrlarano:
> I think we should put these details in the `URL#hash` instead, that can't collide with client solutions in the same way.
> 
> I know this proposal is opt-in so nothing breaking off the bat, but if integrators don't test it properly they might find themselves with a semi-broken solution. Imagine a JSP application that errors on unexpected query parameters, then we add some of ours client side and only when the user tries to refresh or shares the link will an error happen.

I see, this is disabled by default though. But I think I get what you mean in your sample. I'll try to convert it to hash :D

lib/kits/core-ui/components/helpers/component.js
```diff
@@ -123,3 +123,25 @@ export const getColorBrightness = (color) => {
...
+
+    queryParams?.forEach(({name, val}) => {
+        urlParams[action](name, val);
+    });
+
+    if (history.pushState) {
+        urlParams = urlParams.toString() === '' ? '' : `?${urlParams}`;
+        const newUrl = `${window.location.protocol}//${window.location.host}${window.location.pathname}${urlParams}`;
+        window.history.pushState({path: newUrl}, '', newUrl);
+    }
```
klarstrup: You can save a lot of this work just by cloning the current URL and mutating it:

```js
export const pushQueryParam = (queryParams = [], action = "set") => {
  const newUrl = new URL(window.location.href);

  queryParams?.forEach(({ name, val }) => {
    newUrl.searchParams[action](name, val);
  });

  window.history.pushState({ path: String(newUrl) }, "", newUrl);
};
```

Also pushState has been around since IE10, I wouldn't bother feature detecting it.

lib/kits/core-ui/components/common/offer-list.js
```diff
@@ -155,15 +166,36 @@ const OfferList = ({publicationType, configs = {}, sgnViewer, template}) => {
...
+                        name: scriptEls.publicationIdParam,
+                        val: configs.id
+                    },
+                    {
+                        name: scriptEls.pageIdParam,
+                        val: pageNum
+                    }
+                ],
+                pageNum ? 'set' : 'delete'
+            );
```
klarstrup: I think this `pushQueryParam` API is a bit clunky, how do you feel about a call signature that would make this part look like this instead:
```js
pushQueryParam({
  [scriptEls.publicationIdParam]: configs.id,
  [scriptEls.pageIdParam]: pageNum || undefined,
});
```

(`undefined` then means delete the param)

klarstrup (approved):
Just API/implementation nitpicks

Hopefully we can find time for making some tests soon, this is slowly becoming something of a complex beast.

klarstrup (approved):
LGTM!

PR #113, opened 2022-04-19
Feat/offer info and minor enhancements

jrlarano:
- Add labels below the 3 icons at the top (Close, Shopping List, Overview)
- When offer is clicked, we need to present information about the offer first with a Add to Shopping List button
- Add 2 icons for going to first page and last page


mortenbo:
@jrlarano Great job. Here's my feedback:

> @klarstrup Will you concentrate on giving feedback to the code too?

## List of Pubs

Screen Shot 2022-04-19 at 08 42 46

Can you make sure to only allow one line so the layout doesn't break?

## Last/First Page

I like your solution for going to last/first page:

Screen Shot 2022-04-19 at 08 43 01

- There's no cursor when hovering them (doesn't feel like links)
- They should only be shown on desktop/large breakpoints (ie. not mobile)
- How do you disable them via data-*?
- They are a bit too big, IMO

## Fix Label Names

Right now, it says `lukke`, which in incorrect in Danish. Can you change it to `Tilbage`?

Screen Shot 2022-04-19 at 08 43 01

## Wrong Fonts

It looks like we're using incorrect fonts. For example, for the placeholder. We need to inherit the font from the website it's placed in (ie. no custom font in the SDK) and make sure we use the same inherited font everywhere.

Screen Shot 2022-04-19 at 08 43 19

jrlarano:
@mortenbo Thanks for the feedback. These are noted. I'll adjust it accordingly :)

Btw, here are the new data-* options to enable/disable these


header labels: `data-component-publication-show-header-labels="true"`
going to last/first page: `data-component-publication-show-header-controls="true"`
offer overview on offer click: `data-component-publication-viewer-offer-click-behavior="overview_modal"`

mortenbo:
@jrlarano Looks great. More feedback:

1) For some reason, the navigation icons get cut off
Screen Shot 2022-04-21 at 07 48 12

2) Does it make more sense to move last + first page icons next to next + previous page icons?
Screen Shot 2022-04-21 at 07 48 22

3) When clicking "go to first page", the fade out animation is too slow, IMO

4) Fallback shopping list sharing

If the native sharing API is not supported, can you keep the share icon but make it a mailto: where the body is the contents of the shopping list?

5) Offer details
Screen Shot 2022-04-21 at 07 50 50

When you open an offer, what's the header info? Is that for the publication? I think that confuses things because the publication date can differ from the offer date.

Also, the `Add to List` button is too "boring" and hidden. The same goes for the buttons in the shopping list. Can you improve buttons overall? Also, make it `Add to Shopping List`

When you click the add to SL button, can you remove the modal automatically?

When I click a hotspot, there's a 1-2 second delay before the modal opens. The modal should open instantly and show a spinner until the offer is ready.

---

Otherwise, nice job!

klarstrup (approved):

PR #116, opened 2022-05-05
[Viewer] Minor styling fixes

jrlarano:
## Feedback to SDK

Mobile:
1. It cuts the picture
2. Under 'Overview' à 'Pages' here, the pages are not displayed - only numbers are shown

General (mobile + web) :
3. 'Download' does not work.
    Mobile phone:
        It also cuts the image in the downloaded version
        It does not retrieve all pages
4. 'Download' should be called 'Download newspaper'

The following must also be corrected - but not necessarily before we go live

Mobile phone:
5. The navigation arrows are not visible in the mobile version.
 
General (mobile + web) :
6. Search: right now you have to search for specific words for the search to appear. For example. here with 'Nutella'
7. Click on offer: by clicking on the offer, the image of the item does not get bigger. The item is pictured with the same size when you see it in the newspaper as when you click on the offer to see more / clearer.

Web:
8. Share: We first shared the inlet list, but got these two bugs. Is it because this is a test version?

### Based on JsFiddle and local version comparison via browserstack (mobile)
1. Related to JsFiddle - the browser's toolbar blocks the bottom part of the page in jsfiddle
2. Related to JsFiddle limitations
3. Related to JsFiddle limitations
4. Can be fixed via `data-translation-keys--publication-viewer-download-button="Download newspaper"`
5. Working as expected. Shall we show them even on mobile?
6. Working as expected - API limitations?
7. Valid - Fixed in this PR
8. Related to JsFiddle limitations

### DIFF
--- a/examples/list-publications-template.html
+++ b/examples/list-publications-template.html
@@ -48,7 +48,7 @@
             data-component-publication-disable-global-scrollbar="false"
             data-component-publication-show-header-labels="true"
             data-component-publication-viewer-offer-click-behavior="overview_modal"
-            data-translation-keys--publication-viewer-download-button="Download newspaper"
+            data-translation-keys--publication-viewer-download-button="Download PDF"
             defer
         ></script>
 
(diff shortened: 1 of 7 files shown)


klarstrup (approved):
LGTM

PR #118, opened 2022-05-10
Migrate all source code to TypeSrcript without affecting output code

klarstrup:
i forgot to "demo" this but i also innovated it

This is the opposite approach from the one in https://github.com/shopgun/squid/pull/2050 - a near-total immediate migration.

Instead of making any logical changes, any place that has an indubitable type(and potential runtime) error is marked out with a `//@ts-expect-error` which can then be fixed in a followup PR.

The aim of this PR is to migrate to typescript without any changes in the final output `sgn-sdk.js`, and it's basically there. 

What's missing is TS versions of `core-ui` components, which I've decided to hold back on until we have the next version of those fleshed out with rema/etc corrections.


mortenbo:
@klarstrup That's a doozy. Nice job 😃 @jrlarano Could you provide your feedback?

jrlarano:
@mortenbo  @klarstrup , apologies, I'm still not so familiar in using TypeScript so I can't say much about these changes. However for me, this will be a great opportunity to learn and practice TypeScript and I think this is really nice😄

PR #122, opened 2022-06-06
[Viewer] Dispatch even on Incito load

jrlarano:


mortenbo:
@jrlarano What's the purpose of this PR? Not a fan of emitting global events to window.

jrlarano:
> @jrlarano What's the purpose of this PR? Not a fan of emitting global events to window.

Hi @mortenbo, this is only for the clients to be able to use the Incito data eg: `table_of_contents` to create their own toc menu without having to perform another api call

mortenbo:
Maybe this should be solved in a more sandboxed way? I don't like emitting global events. Can't they call a function via SGN to get the table of contents?

jrlarano:
Hmm I'm still not sure if they can call a function in SDK to get the current Incito data without the need to instantiate it. But I will try :)

t-lansing:
Awesome 🤩

Works as expected.

.babelrc
```diff
@@ -23,5 +23,5 @@
             }
         ]
     ],
-    "plugins": [["@babel/plugin-transform-runtime", {"corejs": 3}]]
+    "plugins": [["@babel/plugin-transform-runtime", {"corejs": 3}], "@babel/plugin-proposal-class-properties"]
```
mortenbo: What does this do?

mortenbo:
Is there a way to simplify even more without specifying manual render but have a hook into the rendering part?

.babelrc
t-lansing: this was related to my local environment and i removed it again.

jrlarano:
Will try it again :)

examples/incito-publication-template.html
```diff
@@ -66,7 +66,16 @@
             data-component-publication-disable-global-scrollbar="false"
             data-translation-keys--publication-viewer-shopping-list-label="Shopping List"
             data-translation-keys--publication-viewer-download-button="Download Now"
-            defer
         ></script>
+
+        <script>
+            (async () => {
+                const publicationData = await SGN.PublicationData
```
mortenbo: Missing ;

mortenbo: I like it - but could SGN.PublicationData be tweaked? Could maybe be `SGN.component.{{componentName}}`?

examples/incito-publication-template.html
```diff
@@ -66,7 +66,16 @@
             data-component-publication-disable-global-scrollbar="false"
             data-translation-keys--publication-viewer-shopping-list-label="Shopping List"
             data-translation-keys--publication-viewer-download-button="Download Now"
-            defer
```
mortenbo, line 69: This is no go, IMO. We need to defer so our SDK doesn't block the entire website.

klarstrup:
Maybe I'm late to the party, my proposed solution would've been dispatching events on the target container element to be consumed like so:

```js
document
    .querySelector('#publication-container')
    .addEventListener('incito:rendered', ({details: data}) => data); // {details: {…}, incito: {…}}
```

mortenbo:
@klarstrup I think that's a good idea. Only thing is that we need to align and nail naming. Instead of incito:rendered then maybe publication:rendered or publication:started (same event for both PDF + Incito).

jrlarano:
@mortenbo @klarstrup Thanks for your inputs! I think both solutions make sense :) Let me know what we're going to proceed 🚀

mortenbo:
@jrlarano Let's do Io's suggestion but change naming.

examples/incito-publication-template.html
```diff
@@ -68,5 +68,13 @@
             data-translation-keys--publication-viewer-download-button="Download Now"
             defer
         ></script>
+
+        <script>
+            document
+                .querySelector('#publication-container')
+                .addEventListener('tjek-publication:rendered', (e) => {
```
mortenbo: Lose the tjek- because we're using the container element here. Plus, we don't use the word Tjek in the SDK right now (SGN).

examples/list-publications-template.html
```diff
@@ -62,7 +62,7 @@
             );
         </script>
 
-        <!--   Sample event listener on Incito offer -->
+        <!--   Sample event listener on Incito click -->
         <script>
             window.addEventListener(
                 'tjek-publication-list-item-clicked',
```
mortenbo: Do we still emit events on window?

examples/list-publications-template.html
```diff
@@ -71,5 +71,14 @@
                 }
             );
         </script>
+
+        <!--   Sample event listener on Publication onload -->
+        <script>
+            document
+                .querySelector('#list-publications')
+                .addEventListener('tjek-publication:rendered', (e) => {
```
mortenbo: Lose tjek-

examples/list-publications-template.html
jrlarano: Yes, this is the old implementation. Will refactor it to use the container element :)

lib/kits/core-ui/paged-publication.js
```diff
@@ -313,6 +314,18 @@ const PagedPublication = (
             })
         )?.[0]?.id;
 
+    const dispatchPublicationData = () => {
+        const mainContainerEl = document.querySelector(
+            scriptEls.listPublicationsContainer || scriptEls.mainContainer
+        );
+
+        mainContainerEl?.dispatchEvent(
+            new CustomEvent('tjek-publication:rendered', {
```
mortenbo: You're still using tjek- here.

jrlarano: Ow sorry I missed this one 😅

lib/kits/core-ui/list-publications.js
```diff
@@ -112,8 +112,8 @@ const ListPublications = (
         });
 
     const dispatchPublicationClickEvent = (detail) => {
-        window.dispatchEvent(
-            new CustomEvent('tjek-publication-list-item-clicked', {
+        document.querySelector(scriptEls.mainContainer)?.dispatchEvent(
+            new CustomEvent('publication_list:clicked', {
```
mortenbo: Maybe publication:clicked - does it matter it's a list?

mortenbo (approved):

lib/kits/core-ui/list-publications.js
jrlarano: Ah yeah, good point. It doesn't matter if it's a list since it's the publication that's clicked. Will rename this :)

klarstrup (approved):
I think this is a good compromise between window/global events/variables and having to call into the SDK or CoreUIKit manually

jrlarano:
Hi @mortenbo , @klarstrup When do you think we can release this? A client needs this to create their own menu.

mortenbo:
All good from my side.

PR #130, opened 2022-07-14
[Viewer]Dispatch event on offer modal rendered

jrlarano:
Dispatch event upon opening the offer modal
Event name: `publication:offer_modal_rendered`

### DIFF
--- a/lib/kits/core-ui/components/common/offer-overview.ts
+++ b/lib/kits/core-ui/components/common/offer-overview.ts
@@ -93,6 +93,7 @@ const loaderTemplate = `\
 const OfferOverview = ({
     template,
     configs,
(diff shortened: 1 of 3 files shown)


mortenbo (commented):
Cool! Is `publication:offer_modal_rendered` the same name for Incito or?

jrlarano:
> Cool! Is `publication:offer_modal_rendered` the same name for Incito or?

Yes this will be for both Incito and PDF. It will only be dispatched when opening an offer modal. 
Incito however will now have two offer events: 
`publication:offer_clicked` and `publication:offer_modal_rendered`

klarstrup (approved):

PR #135, opened 2022-08-24
PagedPublicationViewer/EventTracker: Track paged publication open on open

klarstrup:
Addresses https://github.com/shopgun/TjekTeams/issues/540 for Vizsla and 3rd party integrators

### DIFF
--- a/lib/kits/events/tracker.ts
+++ b/lib/kits/events/tracker.ts
@@ -116,7 +116,7 @@ class Tracker {
         return this.trackEvent(1, properties, version);
     }
 
-    trackPagedPublicationPageDisappeared(properties, version?: number) {
+    trackPagedPublicationPageOpened(properties, version?: number) {
         return this.trackEvent(2, properties, version);
     }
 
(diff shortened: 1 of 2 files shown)


lib/kits/events/tracker.ts
```diff
@@ -116,7 +116,7 @@ class Tracker {
         return this.trackEvent(1, properties, version);
     }
 
-    trackPagedPublicationPageDisappeared(properties, version?: number) {
+    trackPagedPublicationPageOpened(properties, version?: number) {
```
klarstrup, line 119: This is actually a breaking change for anyone who has manually been calling `eventTracker. trackPagedPublicationPageDisappeared()` directly instead of letting the `PagedPublicationViewer` do it for them - but I don't believe we have integrators matching that profile at all.

but it was hurting my soul that the name of this method was grossly inaccurate to the Wolf 2.0 event 2 spec name

lib/kits/paged-publication/event-tracking.ts
```diff
@@ -36,11 +36,11 @@ class PagedPublicationEventTracking extends MicroEvent {
         return this;
     }
 
-    trackPageSpreadDisappeared(pageNumbers: number[]) {
+    trackPageSpreadAppeared(pageNumbers: number[]) {
```
mortenbo, line 39: Could also just call it trackPageSpread and alias trackPageSpreadDisappeared to trackPageSpread.

PR #141, opened 2022-10-20
Page Decoration

### LINKED ISSUE #140: [PagedPublicationKit] Page links
- Parent issue: https://github.com/tjek/TjekTeams/issues/628

- Prerequisite issue:  https://github.com/tjek/vizsla/issues/348

---

A new endpoint `/v2/catalogs/[id]/page_decorations` with the output shape `[{page_number, title, website_link}]` is to be used to fetch links to turn certain pages into per-page links.

jrlarano:


lib/kits/core-ui/page-decoration.ts
```diff
@@ -0,0 +1,64 @@
...
+    template?: string;
+    x: number;
+    y: number;
+    pageDecoration: V2PageDecoration;
+}) => {
+    const render = () => {
+        const el = document.createElement('div');
+
+        el.className = 'sgn-pagedecoration';
+        el.setAttribute('tabindex', '-1');
```
klarstrup: this shouldn't be necessary for a `<div>`, but do make sure that the primary `<a>` link is accessible via keyboard navigation.

lib/kits/paged-publication/bootstrapper.ts
```diff
@@ -18,7 +18,11 @@ export default class Bootstrapper {
     }
 
     createViewer(
-        data: {details: V2Catalog; pages: V2Page[]},
+        data: {
+            details: V2Catalog;
+            pages: V2Page[];
+            pageDecorations: V2PageDecoration[];
```
klarstrup: This should be optional, as someone might intentionally want to start a viewer without decorations(for example to implementation their own link decorations outside of PagedPublicationKit, like eta.dk might do)

klarstrup:
While the `/v2/[catalog_id]/page_decorations` endpoint will never be as slow as `/v2/[catalog_id]/hotspots`, I think it should not be allowed to block the main `Bootstrapper.fetch -> Bootstrapper.createViewer`. In my opinion it is fine if page links "pop in" similarly to how hotspots do(and the similar announcement bubble in the main apps).

It would be nice if page decoration loading and display is deferred until after the main viewer start, similarly to the existing `Bootstrapper.fetchHotspots` and `Bootstrapper. applyHotspots` methods and their usage today.

lib/kits/core-ui/components/paged-publication/page-decoration.ts
```diff
@@ -0,0 +1,61 @@
...
+        return el;
+    };
+
+    const getHostname = (link = '') => {
+        try {
+            const url = new URL(link);
+
+            return url.hostname;
+        } catch (e) {
+            return '';
```
klarstrup: This needs a better fallback, if a `website_link` is invalid log an error to the console and don't render the decoration at all.

lib/kits/paged-publication/viewer.ts
```diff
@@ -310,6 +311,28 @@ class Viewer extends MicroEvent {
         this.processHotspotQueue();
     }
 
+    applyPageDecorations(pageDecorations) {
```
klarstrup, line 315: This method should follow the behavior of `Viewer#applyHotspots` and render the decorations into the viewer.

If more advanced integrators(for example our SDK's `paged-publication` component) want to render the decorations outside of the viewer, they should simply not use `Bootstrapper#applyPageDecorations` and instead use the result of `Bootstrapper#fetchPageDecorations` to render their page decorations as they see fit.

klarstrup (changes requested):
Some comments about the structure and behavior of this feature

klarstrup (approved):

PR #151, opened 2022-12-30
bugfix: Incito video loading

mortenbo:
This PR fixes two critical things:

1) If your phone is on low energy mode, the first frame of the video will now be shown as a poster meaning you'll see content
2) We had a major bug in our lazyloading meaning we wouldn't unobserve the video element causing it to re-load over and over again when scrolling to it


mortenbo:
@klarstrup I would like this PR to be included with the other merged PR where we use native lazyloading.

t-lansing (approved):

klarstrup (approved):
LGTM, FFTM!

PR #153, opened 2023-01-02
Remove use of CloudFront so we're cloud agnostic

mortenbo:
Everything is set up in AWS so the domain is good to go.

Once this is merged, please ping @Jakob-tjek so that he knows. We don't need clients to update the domain yet but going forward, all must use it + we should start to get clients to upgrade to latest version anyways.


klarstrup (approved):
LGTM, FFTM!

PR #154, opened 2023-01-03
Reduce delay to 0 if reduce motion is preferred

mortenbo:


lib/verso-browser/verso.ts
```diff
@@ -716,6 +734,10 @@ calls .first()/.prev()/.next()/.last()/.navigateTo() on the viewer.
         return false;
     };
 
+    onPrefersReducedMotionChange = (e) => {
+        this.prefersReducedMotion = !!e.matches;
```
klarstrup: `Boolean(e.matches)`

klarstrup (approved):
Nice!

lib/verso-browser/verso.ts
mortenbo: Fixed.

PR #155, opened 2023-01-05
Simplify video autoload for Incito

mortenbo:
Seeing some of these in Chrome: `Uncaught (in promise) DOMException: The play() request was interrupted by a new load request. https://goo.gl/LdLk22`

This should fix it.

### DIFF
--- a/lib/incito-browser/incito.ts
+++ b/lib/incito-browser/incito.ts
@@ -788,7 +788,9 @@ export default class Incito extends MicroEvent {
             sourceEl.setAttribute('type', el.dataset.mime);
 
             el.appendChild(sourceEl);
-            el.load();
+
+            // Check if media isn't already being loaded.
+            if (el.networkState !== 2) el.load();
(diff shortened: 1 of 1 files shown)


klarstrup (approved):
LGTM i think 🤔

lib/incito-browser/incito.ts
```diff
@@ -827,30 +827,22 @@ export default class Incito extends MicroEvent {
...
+                    if (entry.target instanceof HTMLVideoElement) {
+                        if (entry.isIntersecting) {
                             entry.target.play();
+                        } else {
+                            // If loading is ongoing, we _have to_ wait for play() to be ready before pausing.
+                            if (entry.target.networkState === 2) {
+                                await entry.target.play();
+                            }
+
+                            entry.target.pause();
```
klarstrup, line 840: note that redundant calls to pause and play are idempotent, i.e. if already playing(or loading) then `play()` is a no-op and if already paused `pause()` is a no-op

PR #156, opened 2023-01-06
[Template]Fix hidden body on print media

jrlarano:


mortenbo (approved):

lib/kits/core-ui/components/common/shopping-list.ts
```diff
@@ -221,8 +221,13 @@ const ShoppingList = ({template}) => {
         container
             ?.querySelector('.sgn-shopping-print-list-btn')
             ?.addEventListener('click', () => {
+                document.body.classList.add('sgn-body-print');
                 window.print();
             });
+
+        window.addEventListener('afterprint', () => {
+            document.body.classList.remove('sgn-body-print');
+        });
```
klarstrup: It's probably negligible effect in reality, but this could end up adding a listener repeatedly(which won't be cleaned up since it's on the window), right?

klarstrup (approved):
LGTM

klarstrup:
This is now live as of shopgun-sdk@4.8.8

mortenbo:
@klarstrup Did you invalidate it?

Screenshot 2023-01-09 at 11 59 04

klarstrup:
No I forgot to answer "yes" to the last step of the publish script after i wrote this 🙃 
image

klarstrup:
see now 
image

mortenbo:
@jrlarano Did it always look like this? It doesn't look good/right.

Screenshot 2023-01-09 at 12 03 15

jrlarano:
@mortenbo Yes, the `Print` button needs to be clicked to display this layout 
image

mortenbo:
@jrlarano OK, that's no good, I'm afraid. Has to work when using cmd+p too.

jrlarano:
@mortenbo Okay, I'll work on that :)

PR #162, opened 2023-01-27
Improve Incito performance with async decoding

mortenbo:
Ref: https://www.js-craft.io/blog/what-does-the-html-image-decoding-async-attribute-do-and-how-can-it-help-us-to-improve-performance/


klarstrup (approved):
Good call. This also fits Incito well because images in incitos don't affect the layout anyway - we know the size of everything long before the image loads.

mortenbo:
@klarstrup Yeah, exactly.

mortenbo:
@martensievers What do you think?

martensievers:
Lgtm 🚀

PR #163, opened 2023-02-07
Improve image fetch prio

mortenbo:


lib/incito-browser/incito.ts
```diff
@@ -235,10 +236,21 @@ function renderView(view, canLazyload) {
             if (isDefinedStr(view.src)) {
                 attrs.loading = 'lazy';
                 attrs.decoding = 'async';
+                attrs.fetchpriority = 'high';
```
klarstrup, line 239: I wonder if this risks having below-the-fold images take priority over images currently-in-viewport images. Did you test this with 3G network emulation or similar?

klarstrup: Maybe we only want to do this for ImageViews that are in the first section of the incito or something to that effect...

klarstrup (commented):
There's a high chance we are slow to the first image simply because clients' websites are maxing out HTTP connections before they get to us. I think `fetchpriority` is the right place to go but I suspect  we'll need to more delicate than setting it on all images

PR #167, opened 2023-02-16
Round numbers for Incito

mortenbo:


klarstrup (changes requested):
Me no comprendo, decimal pixel values are valid 🤔

klarstrup:
If this somehow addresses a11y support issues we were talking about: Can we just do this on the incito generation side for problematic incitos instead? Then they retain control and the fix is out instantly rather than needing to propagate through the SDK

mortenbo:
@klarstrup OK, reverted it now. They'll fix on Incito. But we would still like to clean the style.

lib/incito-browser/incito.ts
```diff
@@ -283,7 +283,9 @@ function renderView(view, canLazyload) {
                     .forEach((style) => {
                         const {0: key, 1: value} = style.trim().split(':');
 
-                        styles[key] = value;
+                        if (key && key.length && value && value.length) {
```
klarstrup, line 294: Don't need this length check, empty strings are falsey by themselves

lib/incito-browser/incito.ts
```diff
@@ -283,7 +283,9 @@ function renderView(view, canLazyload) {
...
+                            styles[key] = value.trim();
```
klarstrup: `value` could be an empty string after the `trim()`, consider trimming before the above `if `clause

klarstrup (approved):
Nits

lib/incito-browser/incito.ts
mortenbo: True - better now?

klarstrup (approved):

mortenbo:
@martensievers Shall we still go with this in the SDK or will you fix it server-side?

PR #170, opened 2023-02-22
perf(browserslist): Drop IE11 support

klarstrup:
# Effective change

Remove aggressive automatic polyfill adding and increase browserslist support level to `supports css-grid`, giving us this base level of browser support: https://caniuse.com/css-grid

# Impact

This reduces the size of our main, largest build targets by half! `sgn-sdk-4.x.x.min.js` going from 

# Background

The `list-publications` component has actually already required CSS Grid support for over a year now, of which we have(to my knowledge) heard any complaints. With this in mind I feel comfortable increasing our minimum browser support limit on the SDK to effectively be: 
- Chrome 57 (Mar 9 2017)
- Safari 10.1 (Mar 27 2017)
- Firefox 54 (Mar 21 2017)
- Safari iOS 10.3 (Mar 27 2017)

# Risks

This is in principle a breaking change, dropping off 0.5% of Danish internet users according to GS: https://gs.statcounter.com/browser-market-share/all/denmark/#monthly-202201-202301

But I don't think we are interested in considering this a breaking change in terms of going from 4.x.x to 5.x.x as that would mean verifying and updating a dozen existing integration. This is also not a breaking change in terms of the APIs offered by the SDK.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -11,20 +11,9 @@
         "setPublicClassFields": true,
         "superIsCallableConstructor": true
     },
-    "targets": {"browsers": ["> 3%", "IE 11"], "node": 10},
-    "presets": [
-        [
-            "@babel/preset-env",
-            {
-                "targets": {"browsers": ["> 3%", "IE 11"], "node": 10},
-                "corejs": 3,
-                "useBuiltIns": "usage",
-                "exclude": ["transform-typeof-symbol"]
-            }
-        ]
-    ],
+    "presets": ["@babel/preset-env"],
     "plugins": [
         "@babel/plugin-syntax-import-assertions",
-        ["@babel/plugin-transform-runtime", {"corejs": 3}]
+        "@babel/plugin-transform-runtime"
     ]
 }
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
     "description": "Tjek SDK for JavaScript.",
     "type": "module",
+    "browserslist": "supports css-grid",
     "main": "dist/shopgun-sdk/sgn-sdk.cjs.js",
     "browser": "dist/shopgun-sdk/sgn-sdk.js",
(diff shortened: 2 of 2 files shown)


mortenbo:
@klarstrup I think this makes sense but a few questions:

1) Is IE11 supported in Vizsla?
2) Have you coordinated with Jakob? Just to make sure before we do it. I'm all for it tho.

klarstrup:
> @klarstrup I think this makes sense but a few questions:
> 
> 1. Is IE11 supported in Vizsla?

Not since May 2021 (https://github.com/tjek/vizsla/issues/90) 

> 2. Have you coordinated with Jakob? Just to make sure before we do it. I'm all for it tho.

I'll request his review

Jakob-tjek:
Where is Microsoft Edge in all this? Isn't that a popular browser?
But I'm fine with removing IE

klarstrup:
> Where is Microsoft Edge in all this? Isn't that a popular browser? But I'm fine with removing IE

There are details in the caniuse link in the OP, but Edge uses the same engine as Chrome since version since version 79 and they share version numbers and features.

mortenbo:
Cool. Let's do it then 🚀 🎸

klarstrup:
Here's a usage-proportional view of the 97% we'll support(green) and what we're leaving on the table(red):
image


These are usage statistics from Sweden, as that seems to be the lowest common denominator among our supported countries.

klarstrup:
Out as of https://github.com/tjek/tjek-js-sdk/releases/tag/release-23.02.2023-13.30

I'm also fixing up the SDK publish flow to include automatically creating GitHub releases documenting which versions were published when and which PRs are included.

mortenbo:
Niiiice 👏

klarstrup:
:> 
image

PR #171, opened 2023-02-23
refactor(core-ui): Do Popover template with preact

klarstrup:
Preact is a rendering library with an API similar to and most of the basic features of React, despite being only slightly larger than Mustache. 

It will save us a lot of messing around with rendering Mustache templates into `innerHTML` and then trying to find the right elements to attach events to with `Gator.on` or whatever. 

It will also ease maintenance and codebase cohesion across Tjek by becoming more similar to Vizsla.

The first commit on this PR is an example of using it in a very local and contained way, to make an example that we can use this quite atomically.


klarstrup:
Next I would hope to test this out on one of the core-ui template components. 

Though this might be a non-starter if we have already committed to supporting custom Mustache templates for all those components, as is supported in the code. Do you know if anyone actually uses that @jrlarano? Who requested that feature?

jrlarano:
@klarstrup this is really cool and clean 💪 If I remember correctly, it was @mortenbo's idea to use templating in every component so that the clients can override our default template and use their custom template.
eg:
https://github.com/tjek/tjek-js-sdk/blob/93aac1301c2f38c3f38423d9e5794ca43773610a/examples/list-publications-template.html#L26

I don't think there's any client that's using it atm but I'm not certain about that.

klarstrup:
Alright, I would like to get rid of them until we have a concrete need - then we can invent something then.

PR #172, opened 2023-02-23
feat(EventTracker): Use sendBeacon instead of Fetch with dispatch on unload rather than on next load.

### LINKED ISSUE #161: Use The Web Beacon API for EventsKit

klarstrup:
Closes #161


lib/kits/events/tracker.ts
```diff
@@ -239,14 +242,14 @@ class Tracker {
         };
 
         if (this.eventsTrackUrl) {
-            dispatch(this.eventsTrackUrl, this.eventsTrackHeaders);
+            this.dispatch();
             this.hasMadeInitialDispatch = true;
```
klarstrup, line 299: This on-instantiation dispatching and tracking of `hasMadeInitialDispatch` we can also get rid of if there aren't events(from a persisted pool) to dispatch immediately when `new Tracker`, but we would want to cover that for a transition for any backed-up event pool for a week or whatever

mortenbo:
@klarstrup I think it looks really good. Let's start out by testing this in Vizsla staging before we release this anywhere else.

mortenbo (approved):

klarstrup:
Then I'd like to see you and @jrlarano merge your approved PRs so those can be released and this can be moved ahead into beta for Vizsla's usage

mortenbo:
I've deleted my PR as it's not relevant after all 👍

jrlarano (approved):

lib/kits/events/tracker.ts
```diff
@@ -351,70 +394,16 @@ class Tracker {
...
-    async (
-        eventsTrackUrl: string,
-        eventsTrackHeaders: Record<string, string>
-    ) => {
-        if (!pool) {
-            console.warn('Tracker: dispatch called with no active event pool.');
-            return;
+    dispatching = false;
+    dispatchLimit = 100;
+    dispatch = throttle(this.dispatchBeacon, 4000);
```
klarstrup: Going to swap this `throttle` for a more suitable `debounce`, such that we don't do a lot of immediate single-event dispatches.

PR #176, opened 2023-03-05
Rejigger IncitoEmbedView to actually POST JSON

mortenbo:


klarstrup:
With respect to https://github.com/tjek/bumblebee/pull/3865 - can we add a `incitoEmbedLoaded` event here? Signature `incitoEmbedLoaded: [{el: HTMLElement}]`. 

Additionally if this is used to load entire sections, we might want to make sure that section visibility is not tracked for sections that are not loaded yet.

klarstrup (changes requested):
With respect to https://github.com/tjek/bumblebee/pull/3865 - can we add a `incitoEmbedLoaded` event here? Signature `incitoEmbedLoaded: [{el: HTMLElement}]`. 

Additionally if this is used to load entire sections, we might want to make sure that section visibility is not tracked for sections that are not loaded yet.

mortenbo:
@klarstrup Thx for the feedback. How about now?

klarstrup (approved):
Lookin' good, excited to finally get this superpower. I suspect there will be snags with the real world integrations other than the scroll-to-offer problem, but let's take 'em as they come.

PR #179, opened 2023-03-13
Replace Rollup with ESBuild

klarstrup:
Much simpler, much faster. 

This is 30(!) times faster to build on my M1 MBP:

`develop @ 541c8d99`:
```
npm run build  53.71s user 6.00s system 201% cpu 29.632 total
npm run build  54.20s user 5.62s system 200% cpu 29.778 total
npm run build  53.89s user 6.07s system 203% cpu 29.534 total
```

VS 

`io/esbuild @ 46f52b2e`:
```
npm run build  1.84s user 0.25s system 218% cpu 0.956 total
npm run build  1.84s user 0.25s system 225% cpu 0.923 total
npm run build  1.90s user 0.25s system 219% cpu 0.981 total
```

Development mode is also just as much faster and now has auto-reloading built-in.


klarstrup:
Browser compatibility probably changed, so that still needs to be triple checked and fixed.

PR #182, opened 2023-03-26
Feat/set lazyload in constructor

martensievers:
To improve the performance and reliability of generating image based previews of an Incito (archive), I would like to be able to manually disable lazy-loading when rendering an Incito preview as that allows me to await `networkIdle` in puppeteer instead of simulating scrolling down the page to trigger image loading.


klarstrup (approved):
I was skeptical of the img lazy change to depend on intersectionobserver support, but img lazy is currently only available on browsers  that have intersectionobserver anyway

mortenbo (approved):
Yeah, what Io said. Not 100 % "clean" but probably alright.

PR #190, opened 2023-04-11
feat(Incito, IncitoPublicationKit): Implement pausing

### LINKED ISSUE #180: [IncitoPublicationKit] Allow for pause()'ing the incito viewer when it is obscured
The main goal would be to provide a way to avoid tracking section viewing when for example the incito is obscure by a modal

It would perhaps do well for such a `pause()` to also disable lazy loading, autoplay and such.

This would be elegant if we implement it as a compliment to `start()` and `destroy()`, I think.

klarstrup:
Currently this will only affect section visibility tracking

Fixes #180


klarstrup:
I'm not like super thrilled with having `Incito#unpause()` instead of just making `Incito#start()` idempotent and letting them call that

mortenbo (approved):

PR #202, opened 2023-05-08
Jl/sidebar enhancement

jrlarano:
Incito sidebar improvements:
- Hover and Active highlight effect
- Reflect sectionid in hash/query param
- Support section deeplink

### DIFF
--- a/lib/kits/core-ui/components/common/header.ts
+++ b/lib/kits/core-ui/components/common/header.ts
@@ -326,7 +326,8 @@ const Header = ({
 
         pushQueryParam({
             [scriptEls.publicationIdParam]: null,
-            [scriptEls.pageIdParam]: null
+            [scriptEls.pageIdParam]: null,
+            [scriptEls.sectionIdParam]: null
         });
         location.hash = '';
     };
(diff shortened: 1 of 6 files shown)


klarstrup (approved):

klarstrup (approved):

jrlarano:
@mortenbo waiting for your feedback/approval as well 🙂

mortenbo:
@jrlarano Don't wait on me - or just poke me on Slack 👍

@klarstrup One question though: Can't we consolidate some of this logic in relation to section tracking?

klarstrup:
@mortenbo Sorry, I don't follow. I don't see anything about event tracking in this PR?

@jrlarano Merge at your leisure

mortenbo:
@klarstrup Not what I meant. I meant this PR has hooks for when a section scrolls into view, which is the exact same logic as the event tracking for sections. Why not have "global" hooks you can listen in on, for this purpose here too?

klarstrup:
> @klarstrup Not what I meant. I meant this PR has hooks for when a section scrolls into view, which is the exact same logic as the event tracking for sections. Why not have "global" hooks you can listen in on, for this purpose here too?

Oh those events aren't useful for this at all, they can't tell you which section is the most visible one and they won't fire reliably for that as two sections can be visible simultaneously, scrolling can happen and which one is the most visible one will change without the event firing again as no section has truly appeared/disappeared

mortenbo:
OK 👍

PR #207, opened 2023-05-25
Jl/catch push state error

jrlarano:


klarstrup:
This just breaks `data-component-publication-display-url-params="query"` instead of crashing in the affected scenarios, is fixing the root cause not possible?

jrlarano:
I believe the root cause is the History API in which it doesn't support manipulating the history state for "about:srcdoc" URLs that we use in iframe. I think this is the only scenario that it's failing, and we only use this in the CMS's Integration page. I'm not sure if there's another solution to this 🤔

klarstrup (approved):

PR #214, opened 2023-07-13
Jl/powered by incito

jrlarano:


lib/kits/incito-publication/bootstrapper.ts
```diff
@@ -77,6 +78,10 @@ export default class Bootstrapper {
             Math.abs(window.orientation) === 90
                 ? Math.min(this.options.el.offsetWidth, screen.width)
                 : this.options.el.offsetWidth;
+
+        if (['AFQp2t', 'zUnx3b', 'x1rYHE'].includes(this.options.apiKey)) {
```
simonmarqvard, line 82: As I understand this block is temporary until Davidsen has implemented their displayBranding = false correctly in their integration, correct?

lib/incito-browser/incito.ts
```diff
@@ -757,6 +764,41 @@ export default class Incito extends MicroEvent<{
...
+            };
+            const breakpoint =
+                Object.keys(breakpoints).find(
+                    (name) =>
+                        (this.incito.root_view?.layout_width || 1200) <=
+                        breakpoints[name]
+                ) || 'lg';
+            const padding = paddings[breakpoint];
+
+            this.incito.root_view?.child_views?.push({
```
simonmarqvard, line 787: @Jevy - this position works regardless of whether there's a outro section or not? I know we have had this as a bug in Bumblebee.

simonmarqvard:
Generally I think this looks good! And I think this is the correct place to integrate `powered by incito`. 
Is there a way for me to test this locally - like a test-integration link? Or can I only to that when it is staging/develop? 

I'm sure Jakob and Jeppe would also have some strong opinions about the visuals

lib/incito-browser/incito.ts
simonmarqvard: @klarstrup @jrlarano . I think you pointed it out here - [here](https://github.com/tjek/bumblebee/issues/4010#issuecomment-1628342337)  that we currently have a similar logic to the above in BB but only for catalogs in the App (using API-keys). But we can currently see a difference between a view in the app vs in vizsla :
 If I open this link from my phone browser `https://etilbudsavis.dk/XL-BYG/tilbudsaviser/Robotplaeneklippere~p:K7wQE4xn`  I do not see the `powered by incito` but if I open it in the e*app then I see it. Sometimes I see a quick glitch of it -> and then it disappears. 
 
Is there something we need to be aware of here to render it correctly everywhere?      
If what I'm trying to say here doesn't make sense. Let me know.

mortenbo:
While this PR is nice (except for hardcoding business ID's in here - we should never do this) I don't think it's the right place for it. We need a generalized solution that works across web, iOS, Android (which is the vision of Incito), so I suggest this gets moved to the config, where you can control this with a boolean.

simonmarqvard:
@mortenbo Okay. I guess we do need to implement in BB as we do now then. Just figured it could have worked nicely with the integration page where people can toggle it on/off like the do with menu's, navigation bar etc. But I see your point.

mortenbo:
@simonmarqvard I agree that would be nice. And it could kind of work like that if you wanted to. I'm just arguing it should be persisted in the config and not as a data-* attribute.

PR #217, opened 2023-09-07
Allow inserting Incito into a nested scroll element

mortenbo:
I need to insert Incito into a scrollable div where the html tag has overflow: hidden. This PR allows this.

### DIFF
--- a/examples/incito-publication.html
+++ b/examples/incito-publication.html
@@ -15,6 +15,20 @@
             type="text/css"
         />
         <style>
+            html {
+                overflow: hidden;
+            }
+
+            .sgn__incito {
+                position: absolute;
+                top: 0;
+                left: 0;
+                right: 0;
+                bottom: 0;
+                overflow-y: auto;
+                overflow-x: hidden;
+            }
+
             body {
                 margin: 0;
                 font-size: 16px;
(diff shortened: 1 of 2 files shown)


martensievers (approved):
I'm not very familiar with the SDK but to me it looks fine.

mortenbo:
@klarstrup Any feedback?

klarstrup (changes requested):
Oh, sorry for neglecting this.

I think instead of an option we should aim to "just work" and find the nearest scroll container, something like.

```ts
function getClosestScrollable(el: HTMLElement) {
  const hasScrollableContent = el.scrollHeight > el.clientHeight;
  const isOverflowHidden = window.getComputedStyle(el).overflowY === "hidden";
  if (hasScrollableContent && !isOverflowHidden) return el;

  if (!el.parentElement) return null;

  return getClosestScrollable(el.parentElement);
}
```

Is the move away from the native HTML `loading="lazy"` related to this change?

mortenbo:
We can do that too. The reason for this PR is because if you place Incito inside a div that takes care of overflow instead of body, loading=lazy doesn't work properly. Normally, loading=lazy means it *lazy* loads ahead of time but in this case, it only loads the images once they enter the viewport (vs. ahead of time).

I really don't want to stop using loading=lazy but this was the best alternative solution.

klarstrup:
Gotcha!

mortenbo:
I won't be needing this after all so closing this. Thanks for giving it a looksie though!

PR #220, opened 2024-01-09
[Spans] Add markdown support

jrlarano:
Relates to: https://github.com/tjek/bumblebee/pull/4199, https://github.com/tjek/octopus/issues/305

### DIFF
--- a/lib/incito-browser/incito.ts
+++ b/lib/incito-browser/incito.ts
@@ -76,20 +76,20 @@ function formatSpans(text: string, spans: NonNullable<TextView['spans']>) {
         if (item.span?.style === 'bold') {
             return (
                 memo +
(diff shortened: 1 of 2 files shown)


jrlarano:
Hi @mortenbo this PR is also required for the markdown support 🙂

mortenbo:
@jrlarano Do we need the <span/> and data-name?

jrlarano:
@mortenbo No, we don't need that anymore. I just missed to remove that one. It's removed now 🙂

klarstrup (changes requested):
You need to write the styles directly on the element instead of wrapping with styling elements. `<strong>` and `<em>` aren't reliably styled the same across browsers

klarstrup (approved):

PR #225, opened 2024-03-07
Jl/page decorations improvements

jrlarano:
Note: 
PageDecorations should be passed on applyHotspots to render its hotspost similar to normal hotspots

Eg:
bootstrapper.applyHotspots(sgnViewer, hotspots, sgnPageDecorations);

### DIFF
--- a/lib/kits/core-ui/page-decorations.styl
+++ b/lib/kits/core-ui/page-decorations.styl
@@ -26,68 +26,6 @@
 
     .sgn-pagedecoration-hidden
         display none
-    
-    
-    .sgn-pagedecoration-hotspot
-        position absolute
-        z-index 99
-        box-shadow unset
-        border 0
-        opacity 1
-
-    .sgn-pagedecoration-hotspot-link
-        text-decoration none
-        font-size 14px
-
-    
-    
-    .sgn-pagedecoration-hotspot-link-label
-        outline 0
-        background-color #fff
-        color #3e3e3e
-        border-radius 12px
-        padding 8px
-        box-shadow rgba(0, 0, 0,  0.4) 0px 5px 16px, rgba(0, 0, 0, 0.2) 0px 0px 1px
-        white-space nowrap
-        overflow hidden
-        text-overflow ellipsis
-        margin-left auto
-        margin-right auto
-        opacity 0
-
-    .sgn-pagedecoration-hotspot-link-content
-        display flex
-        align-items flex-end
-        padding 0.5rem
-        
-        &:hover
-            background-color rgba(#fff, 0.1)
-            box-shadow 0 0 15px rgba(#000, 0.3)
-            border-radius 4px
-            border 2px solid rgba(#fff, 0.8)
-
-            .sgn-pagedecoration-hotspot-link-label
-                animation-name sgn-animate-pagedecoration
-                animation-timing-function ease
-                animation-duration .1s
-                opacity 1
-
-    .sgn-pagedecoration-hotspot-link-icon
-        position absolute
-        right 4px
-        top 4px
-        color #3e3e3e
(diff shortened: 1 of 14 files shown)


lib/kits/core-ui/page-decorations.ts
```diff
@@ -42,14 +70,92 @@ const PageDecorations = () => {
...
+
+                    const versoPageSpread = versoPageSpreads?.find(
+                        (pageSpread) =>
+                            pageSpread.pageIds.includes(
+                                `page${pageDecoration.page_number}`
+                            )
+                    );
+                    const contentRect = versoPageSpread?.getContentRect();
+
+                    const pageSpreadEl = pageSpreads
```
mortenbo: Couldn't we make it so that we simply add page decorations to the hotspots array before rendering? Then we kind of skip this code path.

lib/kits/core-ui/page-decorations.styl
```diff
@@ -26,6 +26,67 @@
 
     .sgn-pagedecoration-hidden
         display none
+    
+    
+    .sgn-pagedecoration-hotspot
```
mortenbo: Could you also unify the CSS with hotspots? Essentially extending hotspots to support links and videos.

lib/kits/paged-publication/bootstrapper.ts
```diff
@@ -44,7 +44,58 @@ export default class Bootstrapper {
...
+    applyHotspots(
+        viewer: Viewer,
+        hotspots: V2Hotspot[],
+        pageDecorations?: V2PageDecoration[]
+    ) {
+        let pageDecorationHotspots;
+
+        if (pageDecorations?.length) {
+            pageDecorationHotspots =
+                this.transformPageDecrationHotspots(pageDecorations);
```
mortenbo: Typo

lib/kits/paged-publication/hotspots.ts
```diff
@@ -45,6 +45,23 @@ function getPosition(pages: Page[], ratio: number, hotspot: V2Hotspot) {
     };
 }
 
+function getHostname(link: string) {
```
mortenbo: I think let's remove the link icon and hostname and instead just write in the top/right corner: "Read more", which is localized and which you can override

jrlarano: Here's how it will look like:
image

klarstrup (approved):

jrlarano:
Hi @mortenbo, can we merge this now?

mortenbo:
It's in prod now but I have one change needed: link and embed_link are now strings. So if embed_link is set, use that. If else link is set, use that.

Screenshot 2024-04-15 at 16 41 31

jrlarano:
@mortenbo here's how it should look like now 🙂 


image

mortenbo:
@jrlarano Super, what about if link?

jrlarano:
Link is like this 🙂 
image

mortenbo:
OK, great. We need to fix this: https://github.com/tjek/tjek-js-sdk/pull/225/files#diff-56e8833b30816a4984798688ef87a98d9e5093e18e1ba083ebe1446d64b54e9bR75. You need to set sandbox="allow-scripts allow-same-origin"

mortenbo:
Otherwise, it's fine to merge and deploy if @klarstrup agrees with the PR.

mortenbo:
@klarstrup Once deployed, I'd like to get this into Vizsla too.

klarstrup (approved):

mortenbo:
Is this a regression?
Screenshot 2024-04-19 at 10 50 35

klarstrup:
no that test just sucks

PR #227, opened 2024-04-05
Add option to disable UTM query param

jrlarano:
Relates to: https://github.com/tjek/octopus/issues/400

### DIFF
--- a/lib/kits/core-ui/components/helpers/component.ts
+++ b/lib/kits/core-ui/components/helpers/component.ts
@@ -252,25 +252,31 @@ const updateQueryParam = (url, paramName, newValue) => {
     return urlObject.toString();
 };
 
-export const tranformWebshopLink = (url) => {
+export const transformWebshopLink = (url) => {
     const scriptEl = document.getElementById('sgn-sdk');
     const dataset = scriptEl?.dataset;
 
     if (url) {
+        const newUrl = new URL(url);
+
         if (dataset?.componentPublicationUtmSource) {
-            url = updateQueryParam(
-                url,
+            newUrl.searchParams.set(
                 'utm_source',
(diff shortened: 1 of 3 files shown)


mortenbo (commented):
It feels a bit opposite deleting utm - vs. not setting it if disabled.

jrlarano:
Yeah, but this is because the webshop_link from offer details coming from Squid always contain UTMs. We only added an option to overwrite its default values and an option to remove it if it's disabled.

mortenbo:
Okay. Why the option to remove the params in the first place?

jrlarano:
It's because one of our clients don't like having UTM params when they redirect to the webshop_link

mortenbo (approved):
I think the real solution is to add a new field in Squid so you have two URLs: with and without UTM and then the SDK uses the one without, but this is fine for now.

mortenbo:
Actually, maybe we should completely remove UTM parameters in general using the SDK instead of it being an option? They are not needed when it's the client themselves integrating.

mortenbo:
Nvm my last comment. We'll make it default in the CMS instead, Jakob informed me.

jrlarano:
@mortenbo I see. Does this mean that we won't need this change now?

mortenbo:
Yes, as long as we in the CMS add disable-utm param as a default param in the integration output.

Jakob-tjek:
There has been some misunderstandings here. This PR is still valid.
But another PR for the CMS is required and will add `"data-component-publication-disable-utm="true"` to the integration code if no custom UTM is set:
image
This will make sure we don't change any existing logic for the integrations out there, but at the same time give clients the option to add this line if they want to disable UTM. And new integrations will not get UTM's because of this extra line in the integration code.

PR #230, opened 2024-04-24
Fix pagedecoration hotspot click event

jrlarano:


mortenbo (approved):

jrlarano:
Hi @klarstrup , we need your help in deploying this 🙂

klarstrup:
`shopgun-sdk@4.19.1` is now live on all channels 😁

@mortenbo Maybe it's about time we get @jrlarano set up with the NPM and AWS creds required to publish the SDK?

mortenbo:
@klarstrup What would it require with a similar deploy approach like Squid? Could maybe be powered by GH Actions fully.

klarstrup:
The sdk publish script has a lot more nuance than tagging a commit, multiple packages and versions that change differently requiring multiple inputs. So i don't think that would be possible.
image

PR #233, opened 2024-05-31
Fix shopping list print layout

jrlarano:
Relates to: https://github.com/tjek/octopus/issues/450


jrlarano:
@klarstrup we also need this to be released 🙏

PR #235, opened 2024-06-06
Disable img lazyload if Incito is lazyloaded

jrlarano:


lib/incito-browser/incito.ts
```diff
@@ -720,6 +725,8 @@ export default class Incito extends MicroEvent<{
             canLazyload !== undefined
                 ? canLazyload
                 : 'IntersectionObserver' in window;
+        this.enableLazyLoading =
+            enableLazyLoading !== undefined ? enableLazyLoading : false;
```
klarstrup: Check if it is a boolean

klarstrup (changes requested):
If I understand the purpose of this: I would remove the options and make this the default behavior specifically for incito embed loads ie. at the `this.renderHtml(res)` call in the embed loading code

lib/incito-browser/incito.ts
```diff
@@ -266,7 +266,7 @@ function renderView(view, canLazyload: boolean) {
             const src = String(new URL(view.src));
 
             if (isDefinedStr(view.src)) {
-                if (canLazyload) {
+                if (canLazyload && !enableLazyLoading) {
```
mortenbo: The naming is confusing. You would think enableLazyLoading would enable lazyloading of images. But the enableLazyLoading relates to the Incito

lib/incito-browser/incito.ts
```diff
@@ -866,6 +866,7 @@ export default class Incito extends MicroEvent<{
                     if (res.status === 200) return res.json();
                 })
                 .then((res) => {
+                    this.canLazyload = false;
```
klarstrup, line 869: Not the solution I had in mind, but I can see the appeal 😁

klarstrup (approved):

PR #236, opened 2024-06-07
fix(Incito): Handle non-body scroll container

klarstrup:


mortenbo:
@jrlarano @klarstrup How can we get rid of getScrollContainer? Can't we make it so that the viewer uses <html> by default but if the SDK uses another scroll container, it has to pass it as a param?

jrlarano:
I'm not sure I get the purpose of this. Is this for removing the global scroll that we can see here?
image


If so, they can add this to their integration code `data-component-publication-disable-global-scrollbar="true"`
https://github.com/tjek/tjek-js-sdk/blob/develop/lib/kits/core-ui/incito-publication.ts#L226...#L233

klarstrup:
No, this is to fix IntersectionObserver lazy loading thresholds because they won't trigger ahead of time if the scroll container is another than the one being observed (body by default)

klarstrup:
> @jrlarano @klarstrup How can we get rid of getScrollContainer? Can't we make it so that the viewer uses by default but if the SDK uses another scroll container, it has to pass it as a param?

Uses what by default? 😅

mortenbo:
I think GitHub stripped my comment. I wrote html tag.

PR #238, opened 2024-06-18
Fix img and vid lazyload issue

jrlarano:


mortenbo:
Can we instead just revert everything until when you made the PR? The real fix is fixing the root el of the intersection observer.

PR #252, opened 2024-09-17
fix(publish): Handle binary diff

klarstrup:


klarstrup:
Live now on all channels https://github.com/tjek/tjek-js-sdk/releases/tag/release-18.09.2024-14.07

PR #260, opened 2024-10-25
Cleanup savings template

jrlarano:
This is to avoid the `{{savings}}` template text from showing up in the modal. The reason why there's `{{savings}}` is because we use this as template text for `comment` component which will be tranformed by bumblebee into `Spar {savingsVal}`. However, the squid returns this as it is.

image

### DIFF
--- a/lib/kits/core-ui/components/common/offer-overview.ts
+++ b/lib/kits/core-ui/components/common/offer-overview.ts
@@ -204,6 +204,7 @@ const OfferOverview = ({
 
             return {
                 ...product,
+                description: product.description?.replace('{{savings}}', ''),
                 link: product.link || offer.webshop_link,
                 price,
                 formattedPrice: formatPrice(price, localeCode, priceCurrency),


mortenbo:
I do not believe the JS SDK should contain business logic like this. This ought to be handled server-side.

jrlarano:
@mortenbo Agree, should this be handled in Squid or BB?

mortenbo:
I'd say BB.

jrlarano:
Seems like it's more complicated than I thought. The `{{savings}}` text is coming from the config transformerMap and If I remember correctly, the use of the `{{savings}}` template string is to make the `Comment`, `PriceHeader` `PriceFooter` component to be dynamic and always equal to the savings value. When the client made a manual change in studio and update their savings, the `Spar {savings}` will also change. And another need for this is to dynamically transform the grouped offers with different savings as it will display `Spar optil {savings}`.


https://github.com/tjek/bumblebee/blob/main/lib/generator/offer/megatron.js#L710...L712




image
image

klarstrup:
Does that mean that this PR is off?

jrlarano:
Yes. I'm closing this now. 🙂

PR #262, opened 2024-11-18
Refactor HTMLView

jrlarano:


mortenbo:
I think this is easier to read: const [key, value] = style.trim().split(/:(.+)/);

mortenbo (approved):

PR #263, opened 2024-11-28
Incito: Add carousel layout

mortenbo:


lib/incito-browser/incito.ts
```diff
@@ -788,13 +793,60 @@ export default class Incito extends MicroEvent<{
 
     start() {
         this.el.addEventListener('click', (e) => {
-            const el = closest(
+            const linkEl = closest(
+                e.target as HTMLElement,
+                '.tjek-incito__view [data-link]'
+            );
+            const carouselEl = closest(
+                e.target as HTMLElement,
```
klarstrup: not really related to this PR since this pattern was here before, but you can add `if (!(e.target instanceof HTMLElement)) return;` at the top of the callback and all of these will be inferred correctly, saving all the casting.

klarstrup (approved):
LGTM

Are these carousel elements regular width from server-side?

lib/incito-browser/incito.ts
mortenbo: Good to know, fixed.

mortenbo:
> LGTM
> 
> Are these carousel elements regular width from server-side?

Yeah, exactly. Then the server is in control.

jrlarano (approved):
🚀

PR #268, opened 2024-12-05
fix(incito): almost respect VideoView autoplay/controls/loop options

t-lansing:
There is an issue with the video template, that some of the videos are registered as lazyload and others not within the same publication. 

E.g. here https://www.stark.gl/brochurer/tilbudsavis

Reagardless of the lazyload detection issue, would it not make sense to respect the attributes set on the video view?

### DIFF
--- a/lib/incito-browser/incito.ts
+++ b/lib/incito-browser/incito.ts
@@ -305,7 +305,17 @@ function renderView(view, canLazyload: boolean, shouldLazyload: boolean) {
                 }
             } else {
                 attrs.src = `${src}#t=0.1`;
-                attrs.controls = '';
+                if (view.autoplay === true) {
+                    attrs['data-autoplay'] = true;
+                }
+
+                if (view.controls === true) {
+                    attrs['controls'] = '';
+                }
+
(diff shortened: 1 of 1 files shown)


lib/incito-browser/incito.ts
```diff
@@ -305,7 +305,17 @@ function renderView(view, canLazyload: boolean, shouldLazyload: boolean) {
...
+                    attrs['data-autoplay'] = true;
+                }
+
+                if (view.controls === true) {
+                    attrs['controls'] = '';
+                }
+
+                if (view.loop === true) {
+                    attrs['loop'] = '';
+                }
```
klarstrup: just move this out of the if statement entirely instead of duplicating it

klarstrup (changes requested):

t-lansing:
@klarstrup I will keep an eye out on clients that don't lazyload, and their video settings, so that their videos don't autoplay.

t-lansing:
@jrlarano can you test this befrie deploy?

jrlarano:
Done testing it and it works as expected 🚀

PR #269, opened 2024-12-09
Incito: Stop propagation when clicking links and carousel arrows

mortenbo:


klarstrup:
Is this clear to deploy? Then I can include it in today's Squid deps updates too

mortenbo:
Yeah.

PR #273, opened 2025-02-26
Jl/accessibility improvements

jrlarano:
Relates to: https://github.com/tjek/octopus/issues/662

### DIFF
--- a/lib/kits/core-ui/components/helpers/component.ts
+++ b/lib/kits/core-ui/components/helpers/component.ts
@@ -495,7 +495,7 @@ export const getTotalQuantityByOffer = (savedOffers, offerId) => {
 export const formatDateForReader = (date) => {
     const dateFormat = new Date(date);
 
-    return new Intl.DateTimeFormat('en-US', {
+    return new Intl.DateTimeFormat(navigator.language || 'en-US', {
         month: 'long',
         day: 'numeric'
     }).format(dateFormat);
--- a/lib/kits/core-ui/components/helpers/component.ts
+++ b/lib/kits/core-ui/components/helpers/component.ts
@@ -491,3 +491,12 @@ export const getTotalQuantityByOffer = (savedOffers, offerId) => {
         return totalQuantity;
     }, 0);
 };
+
+export const formatDateForReader = (date) => {
+    const dateFormat = new Date(date);
+
+    return new Intl.DateTimeFormat('en-US', {
+        month: 'long',
+        day: 'numeric'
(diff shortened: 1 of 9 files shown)


lib/kits/core-ui/components/helpers/component.ts
```diff
@@ -491,3 +491,12 @@ export const getTotalQuantityByOffer = (savedOffers, offerId) => {
         return totalQuantity;
     }, 0);
 };
+
+export const formatDateForReader = (date) => {
+    const dateFormat = new Date(date);
+
+    return new Intl.DateTimeFormat('en-US', {
```
mortenbo: Should use navigator.language

lib/kits/paged-publication/viewer.ts
```diff
@@ -279,33 +299,41 @@ class Viewer extends MicroEvent {
...
-            for (const hotspotId in this.hotspots) {
-                if (hotspots[hotspotId]) continue;
 
-                const {id, type, locations, link, embed_link, rotate} =
-                    this.hotspots[hotspotId];
-                for (let idx = 0; idx < hotspotRequest.pages.length; idx++) {
-                    const {pageNumber} = hotspotRequest.pages[idx];
+            hotspotRequest.pages.forEach(({pageNumber}) => {
+                for (const hotspotId in this.hotspots) {
+                    if (hotspots[hotspotId]) continue;
```
mortenbo: Always use curly brackets.

mortenbo (approved):
What about a11y on the images in the pub/hotspots?

jrlarano:
> What about a11y on the images in the pub/hotspots?

Not sure I fully understand what you mean. We only include offer name and price. Do you think we should also include the image offer image in aria label? Won't that be redundant as it will have the same label as the offer name?

mortenbo:
> Not sure I fully understand what you mean. We only include offer name and price. Do you think we should also include the image offer image in aria label? Won't that be redundant as it will have the same label as the offer name?

@jrlarano I just mean if we need aria-labels on the hotspots for easy navigation.

jrlarano:
Hi @mortenbo, ah yes, I've added that here 🙂 

https://github.com/tjek/tjek-js-sdk/blob/cd2631084b659c42a1b856d75a5f87d9d5dc90fe/lib/kits/paged-publication/hotspots.ts#L101

klarstrup (approved):

mortenbo:
@jrlarano Great!

PR #274, opened 2025-02-26
[Refactor] Support Incito3 offer click

jrlarano:


lib/kits/core-ui/components/common/offer-overview.ts
```diff
@@ -202,11 +202,25 @@ const OfferOverview = ({
             const priceCurrency =
                 offer.currency_code || offer.pricing?.currency || currency;
 
+            const totalQuantityByOffer = getTotalQuantityByOffer(
+                storedPublicationOffers,
+                product.id
+            );
+
+            console.log('offer product:', offer, product);
```
mortenbo: Debug

lib/kits/core-ui/components/common/offer-overview.ts
```diff
@@ -224,7 +238,6 @@ const OfferOverview = ({
         const {localeCode, currency, priceFrom} = translations;
         const {offer: incitoOffer} = await fetchOffer({viewId, publicationId});
         offer = incitoOffer;
-        offer.products = transformProducts(offer, products);
```
mortenbo, line 227: Not needed anymore?

lib/kits/core-ui/components/helpers/component.ts
```diff
@@ -426,7 +426,9 @@ export const closeSidebar = () => {
 export const displayOfferMessage = (viewId, message) => {
     if (!message) return;
 
-    const offerContainer = document.querySelector(`[data-id="${viewId}"]`);
+    const offerContainer = document.querySelector(
```
mortenbo: Don't understand this. Why any 3.0 specific changes?

lib/kits/core-ui/components/common/offer-overview.ts
jrlarano: Yes, it will be set in the if-statement just below this line 🙂

mortenbo: Only if >1 (not >0 as I would assume).

lib/kits/core-ui/components/helpers/component.ts
jrlarano: This is because in Incito 3, there are 2 containers with similar data-id. The first container is hidden and the second is the real container. However, due to hierarchy, the hidden one is the one being selected by querySelector so I just added an extra rule to avoid selecting the hidden one.

eg:
image

jrlarano:
Hi @mortenbo , other reason for this adjustment is because some of the current logics are dependent on meta-data `tjek.offer.v1` (eg: products, links) which are no longer available in Incito 3.0

lib/kits/core-ui/components/common/offer-overview.ts
jrlarano: Ah yes, otherwise, it will be set in else statement: https://github.com/tjek/tjek-js-sdk/blob/a474b80fa54aa9eca1eb8e1eab62a1b8819d1e82/lib/kits/core-ui/components/common/offer-overview.ts#L244

Sorry, I was supposed to type if-else statement 😅

klarstrup (approved):

lib/kits/core-ui/components/helpers/component.ts
mortenbo: Wouldn't you instead do clickedEl.closest('[data-role=offer][data-id]')? This is more stable.

jrlarano: Yes! Thanks for the idea! 😄 
Here 🙂 : https://github.com/tjek/tjek-js-sdk/pull/274/commits/ba46a73058c678c3d04e8ca11edf0e54436b5a81

mortenbo (approved):

jrlarano:
Hi @klarstrup, when you have time, can you help us deploy the develop branch (this change along with this PR (https://github.com/tjek/tjek-js-sdk/pull/273)) to production? Thank you! 🙂

